PR #25, opened 2026-02-21
Add FLAG_GRANT_READ_URI_PERMISSION to contact share intent

essentialols:
## Summary
- Adds `FLAG_GRANT_READ_URI_PERMISSION` to the share intent in `shareContact()` so that receiving apps can read the vCard data without needing `READ_CONTACTS` permission

## Problem
When sharing a contact via the share menu, the intent carries a `content://com.android.contacts/contacts/as_vcard/` URI but does not explicitly grant read permission to the receiving app. Apps that lack `READ_CONTACTS` permission (such as Signal, which many GrapheneOS users configure with limited permissions or Contact Scopes) cannot read the vCard data from the content provider URI. This causes the receiving app to report the shared file as invalid.

Sharing via SMS/MMS works fine because those apps typically have broader permissions or handle the URI differently.

## Fix
One-line addition of `intent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)` before the chooser is created. This grants the receiving app temporary read access to the specific vCard URI, regardless of whether it holds `READ_CONTACTS` permission.

While `Intent.createChooser()` is supposed to propagate URI permissions via `ClipData`, this does not reliably work for all URI schemes and permission configurations, particularly with GrapheneOS's Contact Scopes feature. The explicit flag ensures correct behavior.

## Related
- Signal Android vCard sharing issues: [signalapp/Signal-Android#12324](https://github.com/signalapp/Signal-Android/issues/12324)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store
--- a/src/com/android/contacts/interactions/ExportDialogFragment.java
+++ b/src/com/android/contacts/interactions/ExportDialogFragment.java
@@ -21,6 +21,7 @@ import androidx.appcompat.app.AlertDialog;
 import android.app.Dialog;
 import android.app.DialogFragment;
 import android.app.FragmentManager;
+import android.content.ClipData;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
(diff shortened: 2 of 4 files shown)


essentialols:
**Updated:** The fix now covers all three contact share paths (not just single-contact share), and explicitly sets `ClipData` in addition to `FLAG_GRANT_READ_URI_PERMISSION`.

`Intent.createChooser()` propagates URI grants via `ClipData`, not `EXTRA_STREAM`, so the flag alone was insufficient. The three paths fixed:

1. `QuickContactActivity.shareContact()` - single contact
2. `DefaultContactBrowseListFragment.shareSelectedContacts()` - multi-contact
3. `ExportDialogFragment.doShareFavoriteContacts()` - favorites

The `ExportProcessor` file-based export path already handles this correctly via `FileProvider`.

PR #33, opened 2026-06-26
Rewrite SimImport to Kotlin + Compose

### LINKED ISSUE #34: Rewrite SIM import to Kotlin + Compose

sdsantos:
Closes #34

**Feature Changes:**

- When there are no selected contacts to import, the import button is disabled instead of hidden.
- Added Select and Deselect All icon buttons to the topbar.
- Split contacts in two lists: to import and already imported.

**Video of the final result:**

https://github.com/user-attachments/assets/59df6b52-3c8a-44c6-9e21-f78be27c1071

**Implementation Notes:**

- Not sure if the `verification-metadata.xml` my changes should be committed.
- `OptIn` were kept at the class and statement level, and not applied globally.
- Accessibility scanner is passing without suggestions.


app/src/test/kotlin/com/android/contacts/sim/SimImportViewModelTest.kt
```diff
@@ -0,0 +1,256 @@
...
+        val account = AccountInfoFactory.build()
+        val contact = SimContactFactory.build()
+        var startSimImportCall: Triple<Int?, List<SimContact>, AccountWithDataSet>? = null
+        val subject = createViewModel(
+            subscriptionId = subscriptionId,
+            loadAccounts = { flowOf(persistentListOf(account)) },
+            loadSimContacts = { flowOf(SimContactsResult(contacts = persistentListOf(contact))) },
+            startSimImport = { a, b, c -> startSimImportCall = Triple(a, b, c) },
+        )
+
```
RankoR, line 495: `advanceUntilIdle` seems to be required here, otherwise `init` collectors will not run, `currentAccount` will be `null` and `startImport` will early-return

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,229 @@
...
+        loadAccounts()
+            .onEach { accounts ->
+                val defaultAccount = getDefaultAccount()
+                    ?.let { AccountInfo.getAccount(accounts, it) }
+
+                _uiState.update { state ->
+                    state.copy(
+                        isLoading = false,
+                        accounts = accounts,
+                        currentAccount = if (accounts.contains(state.currentAccount)) {
```
RankoR: `AccountInfo` doesn't have `equals` and `hashCode`, so selected account silently resets to default on every accounts reload.

Also, not strictly required, but `when` would look better here (and in many similar cases)

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,229 @@
...
+
+        // Build map of selected contacts per account
+        combine(
+            uiState.map { it.accounts }.distinctUntilChanged(),
+            simContacts,
+        ) { accounts, contacts ->
+            selectedContacts.update { oldSelectedContactsMap ->
+                accounts.associate { account ->
+                    val oldSelectedContacts = oldSelectedContactsMap[account.account]
+                    val selectedContacts = if (oldSelectedContacts.isNullOrEmpty()) {
```
RankoR: `isNullOrEmpty()` conflates "account never initialized" (null) with *"user deselected everything"* (empty set). After Deselect All, the next re-emission - guaranteed by broken `distinctUntilChanged` on any accounts broadcast - re-selects **every contact**. A *partial* selection survives correctly; only the user's most explicit "import nothing" intent is destroyed. The old Java distinguished `ids == null` from an empty `long[]`.

src/com/android/contacts/domain/sim/usecase/LoadSimContacts.kt
```diff
@@ -0,0 +1,49 @@
...
+
+internal class LoadSimContactsImpl @Inject constructor(
+    private val buildBroadcastReceiverFlow: BuildBroadcastReceiverFlow,
+    private val simContactDao: SimContactDao,
+    @param:SimReadDispatcher private val coroutineDispatcher: CoroutineDispatcher,
+) : LoadSimContacts {
+
+    override operator fun invoke(subscriptionId: Int): Flow<SimContactsResult> =
+        buildBroadcastReceiverFlow(IntentFilter(AccountTypeManager.BROADCAST_ACCOUNTS_CHANGED))
+            .onStart { emit(Unit) }
```
RankoR: Make sure to have `catch` downstream, otherwise any exception will crash the app.

src/com/android/contacts/model/account/AccountInfo.java
```diff
@@ -29,6 +31,7 @@
 /**
  * Holds an {@link AccountWithDataSet} and the corresponding {@link AccountType} for an account.
  */
+@Immutable
```
RankoR: It's not immutable. We should probably implement a separate UI-layer model (and a mapper, see examples in Messaging), that will be truly immutable.

src/com/android/contacts/ui/simimport/screen/SimImportScreen.kt
```diff
@@ -0,0 +1,355 @@
...
+@Composable
+internal fun SimImportScreen(
+    effectHandler: SimImportEffectHandler,
+    modifier: Modifier = Modifier,
+    screenModel: SimImportScreenModel = viewModel<SimImportViewModel>(),
+) {
+    val uiState by screenModel.uiState.collectAsStateWithLifecycle()
+    val effectHandler by rememberUpdatedState(effectHandler)
+
+    LaunchedEffect(screenModel, effectHandler) {
```
RankoR: Do not key on `effectHandler`, it will restart collection on change and may drop events. `LaunchedEffect(screenModel)` is enough

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,229 @@
...
+    override val uiState: StateFlow<State> = _uiState.asStateFlow()
+
+    private val simContacts =
+        MutableStateFlow<ImmutableList<SimContact>>(persistentListOf())
+    private val existingContacts =
+        MutableStateFlow<ImmutableMap<AccountWithDataSet, ImmutableSet<Int>>>(persistentMapOf())
+    private val selectedContacts =
+        MutableStateFlow<ImmutableMap<AccountWithDataSet, ImmutableSet<Int>>>(persistentMapOf())
+
+    private val subscriptionId: Int = requireNotNull(
```
RankoR: Don't use `requireNotNull` in production code

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,229 @@
...
+    loadAccounts: LoadAccounts,
+    private val startSimImport: StartSimImport,
+) : ViewModel(),
+    SimImportScreenModel {
+
+    private val _effects = MutableSharedFlow<Effect>(extraBufferCapacity = 1)
+    override val effects: Flow<Effect> = _effects.asSharedFlow()
+
+    private val _uiState = MutableStateFlow(State())
+    override val uiState: StateFlow<State> = _uiState.asStateFlow()
```
RankoR: `StateFlow<State>` type is redundant when you do `asStateFlow()` (check other places too).

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,229 @@
...
+    loadSimContacts: LoadSimContacts,
+    loadAccounts: LoadAccounts,
+    private val startSimImport: StartSimImport,
+) : ViewModel(),
+    SimImportScreenModel {
+
+    private val _effects = MutableSharedFlow<Effect>(extraBufferCapacity = 1)
+    override val effects: Flow<Effect> = _effects.asSharedFlow()
+
+    private val _uiState = MutableStateFlow(State())
```
RankoR: Selection + chosen account no longer survive process death. Old fragment persisted per-account checked IDs and `KEY_SELECTED_ACCOUNT`

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,229 @@
...
+            .launchIn(viewModelScope)
+
+        loadAccounts()
+            .onEach { accounts ->
+                val defaultAccount = getDefaultAccount()
+                    ?.let { AccountInfo.getAccount(accounts, it) }
+
+                _uiState.update { state ->
+                    state.copy(
+                        isLoading = false,
```
RankoR: `isLoading` is cleared on the **accounts** emission, not the SIM read (which is slower). This can cause "No contacts on your SIM card" flash before the list appears.
With zero insertable accounts `currentAccount` stays null, line 141 bails forever, and SIM contacts are never displayed. Old code waited on `Futures.allAsList(accounts, simContacts)`

src/com/android/contacts/ui/core/AppScaffold.kt
```diff
@@ -0,0 +1,29 @@
...
+import androidx.compose.material3.Scaffold
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalLayoutDirection
+
+@Composable
+internal fun AppScaffold(
+    content: @Composable () -> Unit,
+) {
+    Scaffold {
```
RankoR: `AppScaffold` applies start/end padding but never consumes insets, and discards top/bottom entirely. `TopAppBar` then re-applies its own horizontal insets → double inset in landscape with 3-button nav. Bottom is hand-rolled in `ContactsList:210`; the empty-state text and spinner get nothing. Four inset strategies in one screen. Messaging has no `AppScaffold` - each screen owns its `Scaffold` and threads `contentPadding` through.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,229 @@
...
+                    .mapValues { (_, set) ->
+                        set.map { it.recordNumber }.distinct().toImmutableSet()
+                    }
+                    .toImmutableMap()
+            }
+            .launchIn(viewModelScope)
+
+        loadAccounts()
+            .onEach { accounts ->
+                val defaultAccount = getDefaultAccount()
```
RankoR: `getDefaultAccount()` runs on Main on every accounts emission. `ContactsPreferences.java:377-386` wraps it in `StrictMode.permitDiskReads()` around a synchronous binder IPC - the platform itself declares it blocking. Old code called it once.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,229 @@
...
+                    }
+                    account.account to selectedContacts.toImmutableSet()
+                }.toImmutableMap()
+            }
+        }
+            .launchIn(viewModelScope)
+
+        // Set SIM contacts to show in the UI according to the current account,
+        // and based on whether they are selected or already imported
+        combine(
```
RankoR: Second `combine` both reads and writes `_uiState` with no `distinctUntilChanged` on `map { it.currentAccount }`. It terminates only because `StateFlow` conflates structurally-equal values - one future model type without `equals` in `contactsToImport` turns this into an infinite main-thread loop.

src/com/android/contacts/ui/simimport/common/SimContactCell.kt
```diff
@@ -0,0 +1,184 @@
...
+            isLast = isLast,
+            isSelected = isSelected,
+        ),
+    ) {
+        Row(
+            Modifier
+                .fillMaxWidth()
+                .run {
+                    if (isSelectable && onClick != null) {
+                        selectable(
```
RankoR: `toggleable` would be better here

src/com/android/contacts/domain/accounts/AccountsBindsModule.kt
```diff
@@ -0,0 +1,28 @@
+package com.android.contacts.domain.accounts
```
RankoR: DI modules should be in `di/$feature/module` package

src/com/android/contacts/ui/simimport/screen/SimImportScreen.kt
```diff
@@ -0,0 +1,355 @@
...
+                contactsToImport = persistentListOf(),
+                contactsAlreadyImported = persistentListOf(),
+            ),
+            onAction = {},
+        )
+    }
+}
+
+@VisibleForTesting
+const val TEST_TAG_SIM_IMPORT_CONTACTS_TO_IMPORT_TITLE = "sim_import_contacts_to_import_title"
```
RankoR: Test tags should be in a separate file and in most cases with `internal` visibility

src/com/android/contacts/ui/simimport/common/SimContactCell.kt
```diff
@@ -0,0 +1,184 @@
...
+import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.tooling.preview.PreviewLightDark
+import androidx.compose.ui.unit.dp
+import com.android.contacts.model.SimContact
+import com.android.contacts.ui.common.model.SelectableItem
+import com.android.contacts.ui.core.ContactsPreviewColumn
+
+@Composable
+internal fun SimContactCell(
```
RankoR, line 25: `modifier: Modifier = Modifier` is missing. Check other Composables too.

src/com/android/contacts/ui/simimport/screen/SimImportScreen.kt
```diff
@@ -0,0 +1,355 @@
...
+                    contentDescription = stringResource(
+                        R.string.sim_import_cancel_content_description,
+                    ),
+                )
+            }
+        },
+        title = {
+            Text(
+                text = if (uiState.selectedContactsCount > 0) {
+                    uiState.selectedContactsCount.toString()
```
RankoR: While it mirrors old behavior, it's not a good UX to display just a number. Should be a plural string res (English-only is enough for now)

src/com/android/contacts/ui/simimport/screen/SimImportEffectHandler.kt
```diff
@@ -0,0 +1,18 @@
...
+internal interface SimImportEffectHandler {
+    fun handle(effect: Effect)
+}
+
+internal class SimImportEffectHandlerImpl(private val activity: Activity) :
+    SimImportEffectHandler {
+
+    override fun handle(effect: Effect) {
+        when (effect) {
+            Effect.Close -> activity.finish()
```
RankoR: `setResult(RESULT_OK)` / `setResult(RESULT_CANCELED)` are lost here

gradle/libs.versions.toml
```diff
@@ -39,16 +41,19 @@ androidx-compose-ui-test-junit4 = { module = "androidx.compose.ui:ui-test-junit4
 androidx-compose-ui-test-manifest = { module = "androidx.compose.ui:ui-test-manifest" }
 androidx-compose-ui-tooling = { module = "androidx.compose.ui:ui-tooling" }
 androidx-compose-ui-tooling-preview = { module = "androidx.compose.ui:ui-tooling-preview" }
+androidx-lifecycle-viewmodel-compose = { module = "androidx.lifecycle:lifecycle-viewmodel-compose" }
```
RankoR: Version isn't covered by the Compose BOM, so version must be set

RankoR (changes requested):
Please also check for unused resources and imports.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
sdsantos: From what I understood, `getDefaultAccount()` already was called on the main thread by the `SimImportFragment`. I kept calling it whenever accounts might change, because we could lose the currently selected account, and we change to the default account that way. Plus, `ContactsPreferences` caches the `mDefaultAccount` value.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
sdsantos: Well caught. I introduced nullable states in `SimImportUiState` and `simContacts` to ensure we can distinguish between the not having received any value yet, and empty lists. The code get's slightly more complex with nullability checks, but it should help avoiding making this issues.

src/com/android/contacts/model/account/AccountInfo.java
sdsantos: Went with 2 domain models (`AndroidModel` and `AccountDisplayModel`) and a `AndroidUIModel`.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
sdsantos: I had copied the approach from somewhere in Messaging, but switched it now to logging a warning about it and falling back to `SimCard.NO_SUBSCRIPTION_ID`.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
sdsantos: Went with the approach I've seen on Messaging of storing that data on the `savedStateHandle` as soon as the data changes in the ViewModel.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
sdsantos: Remove the `isLoading` argument to avoid conflicting states, and it's now a value derived from the values being non-null.

src/com/android/contacts/ui/core/AppScaffold.kt
sdsantos: When with a plain screen `Scaffold` and passing the `contentPadding` along.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
sdsantos: I update this code to only call `getDefaultAccount()` if we don't have any currentAccount set, or if it's no longer valid, or we don't have anything on `savedStateHandle`. That way, we only call it when strickly necessary, but still covering the scenario where the old default account could have been deleted in the meantime and changed to a new one.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
sdsantos: Now the accounts are kotlin data classes that can be compared correctly.

There's still a theoretical danger of an infinite loop, but the code would have to be changed to actually update the `currentAccount` inside the state update, which is a more obvious issue.

src/com/android/contacts/domain/accounts/mapper/AccountDisplayModelMapper.kt
```diff
@@ -0,0 +1,22 @@
+package com.android.contacts.domain.accounts.mapper
+
+import com.android.contacts.domain.accounts.model.AccountDisplayModel
+import com.android.contacts.domain.accounts.model.AccountModel
```
m4pl: Unused import

src/com/android/contacts/interactions/ImportDialogFragment.java
```diff
@@ -39,16 +38,16 @@
...
 import com.android.contacts.compat.PhoneNumberUtilsCompat;
 import com.android.contacts.database.SimContactDao;
 import com.android.contacts.editor.SelectAccountDialogFragment;
 import com.android.contacts.model.AccountTypeManager;
 import com.android.contacts.model.SimCard;
-import com.android.contacts.model.SimContact;
 import com.android.contacts.model.account.AccountInfo;
 import com.android.contacts.model.account.AccountWithDataSet;
+import com.android.contacts.ui.UIIntents;
+import com.android.contacts.ui.simimport.SimImportActivity;
```
m4pl: Unused import

src/com/android/contacts/ui/simimport/common/SimContactCell.kt
```diff
@@ -0,0 +1,173 @@
...
+                    checked = isSelected,
+                    onCheckedChange = null,
+                )
+            }
+        }
+    }
+}
+
+@Composable
+private fun itemClipShape(
```
m4pl, line 116: Corner radius changes with selection, so a checked row in the middle of the list becomes fully rounded while the rows above and below stay square, leaving odd notches (screenshot). The shape of a row shouldn't depend on whether it is checked.

Image

src/com/android/contacts/ui/simimport/common/SimImportAccountPicker.kt
```diff
@@ -0,0 +1,180 @@
...
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+private fun ExposedDropdownMenuBoxScope.AccountTextField(
+    current: AccountUiModel,
+    canChangeAccounts: Boolean,
+    isExpanded: Boolean,
+) {
+    val currentAccountLabel = current.name.orEmpty()
+    OutlinedTextField(
```
m4pl, line 100: The account name behaves like editable text: long pressing it selects the text and opens the copy/share toolbar. I think it shouldn't be selectable.

src/com/android/contacts/ui/simimport/screen/SimImportScreen.kt
```diff
@@ -0,0 +1,357 @@
...
+            IconButton(onClick = { onAction(Action.CloseClicked) }) {
+                Icon(
+                    imageVector = Icons.Default.Close,
+                    contentDescription =
+                        stringResource(R.string.sim_import_cancel_content_description),
+                )
+            }
+        },
+        title = {
+            Text(
```
m4pl: Title wraps to two lines on small screens.

src/com/android/contacts/domain/sim/usecase/LoadSimContacts.kt
```diff
@@ -0,0 +1,62 @@
...
+
+internal class LoadSimContactsImpl @Inject constructor(
+    private val buildBroadcastReceiverFlow: BuildBroadcastReceiverFlow,
+    private val simContactDao: SimContactDao,
+    private val accountModelMapper: AccountModelMapper,
+    @param:SimReadDispatcher private val coroutineDispatcher: CoroutineDispatcher,
+) : LoadSimContacts {
+
+    override operator fun invoke(subscriptionId: Int): Flow<SimContactsResult> =
+        buildBroadcastReceiverFlow(IntentFilter(AccountTypeManager.BROADCAST_ACCOUNTS_CHANGED))
```
m4pl, line 31: If the SIM is removed while this screen is open, the list keeps showing the SIM contacts and Import stays available. The screen should react to the SIM being removed.

src/com/android/contacts/ui/simimport/screen/SimImportScreen.kt
```diff
@@ -0,0 +1,357 @@
...
+                            R.plurals.sim_import_selected_contacts_title,
+                            uiState.selectedContactsCount,
+                            uiState.selectedContactsCount,
+                        )
+                    else -> stringResource(R.string.sim_import_title_none_selected)
+                },
+            )
+        },
+        actions = {
+            IconButton(
```
m4pl: Two separate buttons where only one is ever enabled. Could this be a single control?

res/layout/fragment_sim_import.xml
```diff
@@ -1,133 +0,0 @@
...
-        android:layout_height="match_parent"
-        android:orientation="vertical">
-
-        <androidx.appcompat.widget.Toolbar
-            android:id="@+id/toolbar"
-            style="@style/LightToolbarStyle"
-            android:layout_width="match_parent"
-            android:layout_height="?attr/actionBarSize"
-            android:elevation="3dp"
-            android:theme="@style/LightToolbarThemeOverlay"
```
m4pl, line 33: These are no longer referenced anywhere after the old SIM import layouts were deleted: `LightToolbarNavigationButtonStyle`, `LightToolbarThemeOverlay`, `LightToolbarStyle`, `FullScreenDialogAnimationStyle`, `PeopleThemeAppCompat.FullScreenDialog`, `PeopleThemeAppCompat.FullScreenDialog.SimImportActivity`, and `res/anim/slide_and_fade_out.xml`.

src/com/android/contacts/ui/simimport/screen/SimImportScreen.kt
m4pl, line 1: Once contacts are imported there's no way to undo it from this screen, you have to delete them one by one in the contacts list. I'd like them to be undoable, but maybe that's not this screen's job.

@RankoR what do you think, is undo in scope here?

src/com/android/contacts/domain/accounts/mapper/AccountDisplayModelMapper.kt
```diff
@@ -0,0 +1,22 @@
...
+import com.android.contacts.model.account.AccountInfo
+import javax.inject.Inject
+
+internal interface AccountDisplayModelMapper {
+    fun map(accountInfo: AccountInfo): AccountDisplayModel
+}
+
+internal class AccountDisplayModelMapperImpl @Inject constructor() : AccountDisplayModelMapper {
+    override fun map(accountInfo: AccountInfo): AccountDisplayModel {
+        return AccountDisplayModel(
```
m4pl, line 29: The account picker shows an empty name on a phone with no Google account, so it's not clear where contacts will be saved (screenshot). `account.name` is `null` for device and SIM accounts, `AccountInfo.nameLabel` has the "Device" / "SIM" label for that case.

Image

src/com/android/contacts/domain/sim/usecase/LoadSimContacts.kt
```diff
@@ -0,0 +1,62 @@
...
+
+                Log.e(TAG, "Load SIM contacts failed (subscriptionId=$subscriptionId)", it)
+                emit(SimContactsResult())
+            }
+            .flowOn(coroutineDispatcher)
+
+    private fun load(subscriptionId: Int): SimContactsResult {
+        val sim = simContactDao.getSimBySubscriptionId(subscriptionId) ?: return SimContactsResult()
+        val contacts = simContactDao.loadContactsForSim(sim).orEmpty()
+        val accountsMap = simContactDao.findAccountsOfExistingSimContacts(contacts).orEmpty()
```
m4pl, line 46: Importing the same SIM contact twice creates a duplicate. Import succeeds, but reopening the screen lists the contact under "Contacts to import" again.

The check matches raw contacts by phone and name. We insert only `StructuredName.DISPLAY_NAME` and the provider re-derives it from the parsed name parts, so `01.Balance` is stored as `01 Balance` and the name comparison in `SimContact.findByPhoneAndName` fails. Verified on device, the query returns the raw contacts and every one comes back as no match. Names without punctuation are fine.

Not caused by this PR, the matching code is untouched. Up to you whether to fix it here or create a separate issue.

m4pl (commented):
Only had time to test this on a device so far, the comments above are what I ran into. Will go through the code properly in the next pass.

src/com/android/contacts/ui/simimport/common/SimContactCell.kt
sdsantos: I followed the Material spec and the latest Google Contacts app: https://m3.material.io/components/lists/specs
But I also feel that the notches were a bit too sharp, so in the next PR, I made all corners have at least 2.dp of corner size. I'll update this PR with the same.

src/com/android/contacts/ui/simimport/common/SimImportAccountPicker.kt
sdsantos: That's the default behavior of readOnly TextFields. Adding a `DisableSelection` around it doesn't fix it.

If we really don't want text selection there, we need to design the custom Button ourselves. Possible, just a bit more work. Is it preferable?

src/com/android/contacts/domain/sim/usecase/LoadSimContacts.kt
sdsantos: It didn't before, but it could start reacting now to the selected SIM card being removed.

I thought it was fine to remove the SIM after loading all the contacts, but it seems the import fails without the SIM.

I imagine we should just close the screen with a message saying the SIM card has been removed?

src/com/android/contacts/ui/simimport/screen/SimImportScreen.kt
sdsantos: Both can be enabled, if you just selected some but not all contacts.

src/com/android/contacts/domain/sim/usecase/LoadSimContacts.kt
sdsantos: Got it. I can reproduce the issue as well.

I feel it should be a separate ticket. For example, importing from a VCard also allows duplicates, and I'm not sure we want that as well there.

sdsantos: https://github.com/GrapheneOS/platform_packages_apps_Contacts/issues/40

src/com/android/contacts/domain/sim/usecase/LoadSimContacts.kt
m4pl: Closing sounds right, but I'd skip the message. In Messaging we just navigate up silently when the opened subscription disappears. The user pulled the SIM out themselves, so they know why. Minor either way.

src/com/android/contacts/ui/simimport/common/SimImportAccountPicker.kt
m4pl: Makes sense, not worth a custom component. Let's leave it.

src/com/android/contacts/ui/simimport/screen/SimImportScreen.kt
m4pl: Got it. Two icons plus the Import button still feels like a lot for the app bar, but it's not a blocker. Fine as is.

sdsantos:
@RankoR rebased to the latest `main`

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,302 @@
...
+    private val selectedContacts = MutableStateFlow(restoreSelectedContacts())
+
+    private val subscriptionId: Int = savedStateHandle[EXTRA_SUBSCRIPTION_ID] ?: run {
+        Log.w(TAG, "$EXTRA_SUBSCRIPTION_ID missing, defaulting to ${SimCard.NO_SUBSCRIPTION_ID}")
+        SimCard.NO_SUBSCRIPTION_ID
+    }
+
+    init {
+        if (subscriptionId != SimCard.NO_SUBSCRIPTION_ID) {
+            loadSimCards()
```
RankoR, line 97: There's no `catch`, so it can crash the app

sdsantos: Done. But the previous and existing code did not catch any exception from the Dao method, and the only mention I could find of a throw was for a `UnsupportedOperationException`, so I added a catch for that one.

RankoR (approved):

src/com/android/contacts/ui/simimport/screen/SimImportScreen.kt
RankoR: I think it's out of scope for this PR

sdsantos:
Rebased to the latest `main`

res/values/strings.xml
```diff
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!-- Copyright (C) 2006 The Android Open Source Project
+<?xml version="1.0" encoding="utf-8"?><!-- Copyright (C) 2006 The Android Open Source Project
```
m4pl: Looks like the line break should be restored.

src/com/android/contacts/domain/accounts/model/AccountDisplayModel.kt
```diff
@@ -0,0 +1,14 @@
...
+import android.graphics.drawable.Drawable
+
+/*
+ * Immutable domain model to match {@link com.android.contacts.model.account.AccountDisplayInfo}
+ */
+internal data class AccountDisplayModel(
+    val account: AccountModel,
+    val name: String?,
+    val type: String? = null,
+    val icon: Drawable? = null,
```
m4pl: `Drawable` can change and is tied to `Resources`, so this model isn't really immutable. It also ends up in `AccountUiModel`, which is marked `@Immutable`, and that's no longer true. Could we keep an id here and build the icon in the UI?

.editorconfig
```diff
@@ -14,11 +14,14 @@ ij_kotlin_allow_trailing_comma_on_call_site = true
 ij_kotlin_name_count_to_use_star_import = 2147483647
 ij_kotlin_name_count_to_use_star_import_for_members = 2147483647
 ij_kotlin_packages_to_use_import_on_demand = unset
+ktlint_class_signature_rule_force_multiline_when_parameter_count_greater_or_equal_than = 1
 ij_kotlin_line_break_after_multiline_when_entry = false
 ktlint_code_style = android_studio
+ktlint_class_signature_rule_force_multiline_when_parameter_count_greater_or_equal_than = 1
```
m4pl, line 19: Duplicate of line 17.

.editorconfig
```diff
@@ -14,11 +14,14 @@ ij_kotlin_allow_trailing_comma_on_call_site = true
...
 ktlint_function_naming_ignore_when_annotated_with = Composable
+ktlint_standard_filename = disabled
 ktlint_standard_function-expression-body = disabled
 ktlint_standard_function-signature = disabled
 ktlint_standard_trailing-comma-on-call-site = disabled
 ktlint_standard_blank-line-between-when-conditions = disabled
-max_line_length = 100
+max_line_length = 100
```
m4pl: Missing final newline.

src/com/android/contacts/ui/simimport/screen/model/SimImportAction.kt
```diff
@@ -0,0 +1,16 @@
...
+internal sealed interface SimImportAction {
+    data object CloseClicked : SimImportAction
+    data class AccountChanged(
+        val account: AccountUiModel,
+    ) : SimImportAction
+    data class ContactSelectionChanged(
+        val contact: SimContactUiModel,
+        val isSelected: Boolean,
+    ) : SimImportAction
+
```
m4pl, line 15: Blank lines inconsistent.

src/com/android/contacts/ui/simimport/screen/model/SimImportUiState.kt
```diff
@@ -0,0 +1,31 @@
...
+import kotlinx.collections.immutable.ImmutableList
+
+@Immutable
+internal data class SimImportUiState(
+    val accounts: ImmutableList<AccountUiModel>? = null,
+    val currentAccount: AccountUiModel? = null,
+    val contactsToImport: ImmutableList<SelectableItem<SimContactUiModel>>? = null,
+    val contactsAlreadyImported: ImmutableList<SimContactUiModel>? = null,
+) {
+    val isLoading
```
m4pl: Looks like `isLoading` stays true forever when `accounts` is empty. `contactsToImport` is only set in the combine at `SimImportViewModel.kt:151`, and that one returns early if `currentAccount` is null, which it always is with no accounts. So the spinner never goes away and the empty-accounts branch below is never reached.

`LoadAccountsImpl.load()` returns an empty list on `ExecutionException`, so this can happen. Could you double check the chain? A test for the empty case would be good too, there's none right now.

src/com/android/contacts/domain/sim/usecase/LoadSimContacts.kt
```diff
@@ -0,0 +1,62 @@
...
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onStart
+
+internal fun interface LoadSimContacts {
+    operator fun invoke(subscriptionId: Int): Flow<SimContactsResult>
+}
+
+internal class LoadSimContactsImpl @Inject constructor(
+    private val buildBroadcastReceiverFlow: BuildBroadcastReceiverFlow,
+    private val simContactDao: SimContactDao,
```
m4pl, line 25: Use cases reach platform code directly here and in `LoadAccountsImpl`, `GetDefaultAccountImpl`, `LoadSimCardsImpl`, `StartSimImportImpl`. It would also make splitting into modules harder later. Not blocking, up to you whether to change it now.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,302 @@
...
+    private fun startImport() {
+        val state = uiState.value
+        val account = state.currentAccount?.account ?: return
+        val selectedContacts = getSelectedContacts() ?: return
+
+        startSimImport(subscriptionId, selectedContacts, account)
+        close(isSuccessful = true)
+    }
+
+    private fun getSelectedContacts(): ImmutableList<SimContact>? {
```
m4pl: `map` then `unmap` is lossless today, but a reverse mapper is easy to forget: add a field to `SimContact` and it silently won't be imported. Keeping the original list around would let `unmap` go. Fine as is though.

Also `?.ifEmpty { null } ?: return` makes an empty selection a silent no-op.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,302 @@
...
+                    state.copy(
+                        accounts = uiAccounts,
+                        currentAccount = findCurrentAccount(uiAccounts),
+                    )
+                }
+            }
+            .launchIn(viewModelScope)
+
+        // Build map of selected contacts per account
+        combine(
```
m4pl: Side effects inside the `combine` transform, same at :151. `uiState` is also an input and an output at once, so the loop only stops because `distinctUntilChanged` filters the fields these blocks write. Would one derived flow work better?

```
override val uiState: StateFlow<State> =
    combine(accounts, simContacts, selectedContacts, existingContacts, pickedAccount) {
        accounts, contacts, selected, existing, picked -> State(...)
    }.stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), State())
```

`WhileSubscribed` would also drop the broadcast receiver and the subscription listener while the screen is off.

src/com/android/contacts/ui/simimport/screen/model/AccountUiModel.kt
```diff
@@ -0,0 +1,32 @@
...
+@Immutable
+internal data class AccountUiModel(
+    val name: String?,
+    val type: String? = null,
+    val icon: Drawable? = null,
+    private val accountName: String? = null,
+    private val accountType: String? = null,
+    private val accountDataSet: String? = null,
+) {
+    constructor(accountDisplayModel: AccountDisplayModel) : this(
```
m4pl: Mapper hidden in a constructor, while `SimContactUiModel` uses an injected one. And `AccountModel` is split into three private fields only to be rebuilt in the getter at :26, was there a reason not to keep `val account: AccountModel`?

src/com/android/contacts/util/core/BuildBroadcastReceiverFlow.kt
m4pl, line 1: Platform wrapper, not domain.

app/src/test/kotlin/com/android/contacts/sim/SimImportViewModelTest.kt
```diff
@@ -0,0 +1,432 @@
...
+        )
+        advanceUntilIdle()
+
+        subject.effects.test {
+            subject.onAction(Action.ImportClicked)
+            advanceUntilIdle()
+            assertEquals(Effect.Close(isSuccessful = true), awaitItem())
+        }
+        assertNotNull(startSimImportCall)
+        startSimImportCall!!.let { (callSubscriptionId, callContacts, callAccount) ->
```
m4pl: Could be `verify { startSimImport(...) }` instead, that drops the `var`, the `Triple`, `assertNotNull` and the `!!`. `SimImportScreenTest` already uses `mockk`.

app/src/test/kotlin/com/android/contacts/sim/SimImportViewModelTest.kt
```diff
@@ -0,0 +1,432 @@
...
+        val account = AccountDisplayModelFactory.build()
+        val contact = SimContactFactory.build()
+        val subject = createViewModel(
+            loadAccounts = { flowOf(persistentListOf(account)) },
+            loadSimContacts = { flowOf(SimContactsResult(contacts = persistentListOf(contact))) },
+        )
+        advanceUntilIdle()
+
+        with(subject.uiState.value) {
+            assertEquals(1, contactsToImport!!.size)
```
m4pl: `contactsToImport!!` appears 14 times here because `SimImportUiState` has four separate nullable fields. A sealed state would make these plain reads, and would fix the `isLoading` issue too.

src/com/android/contacts/domain/sim/usecase/LoadSimContacts.kt
```diff
@@ -0,0 +1,62 @@
...
+    private val accountModelMapper: AccountModelMapper,
+    @param:SimReadDispatcher private val coroutineDispatcher: CoroutineDispatcher,
+) : LoadSimContacts {
+
+    override operator fun invoke(subscriptionId: Int): Flow<SimContactsResult> =
+        buildBroadcastReceiverFlow(IntentFilter(AccountTypeManager.BROADCAST_ACCOUNTS_CHANGED))
+            .onStart { emit(Unit) }
```
m4pl: `onStart` runs before the receiver is registered, so broadcasts in that window are lost. Same at `LoadAccounts.kt:37`. Moving the initial `trySend(Unit)` into `BuildBroadcastReceiverFlowImpl` fixes both.

src/com/android/contacts/domain/sim/usecase/StartSimImport.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.android.contacts.model.SimContact
+import com.android.contacts.model.account.AccountWithDataSet
+import dagger.hilt.android.qualifiers.ApplicationContext
+import javax.inject.Inject
+import kotlinx.collections.immutable.ImmutableList
+
+internal fun interface StartSimImport {
+    operator fun invoke(
+        subscriptionId: Int,
+        contacts: ImmutableList<SimContact>,
```
m4pl: List would be enough, this never reaches Compose.

app/src/test/kotlin/com/android/contacts/tests/SimContactFactory.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import com.android.contacts.model.SimContact
+import kotlin.random.Random
+
+object SimContactFactory {
+    fun build(
+        recordNumber: Int = Random.nextInt(),
+        name: String = "Contact Name",
+        phone: String? = null,
+
```
m4pl: Stray blank line here.

src/com/android/contacts/ui/simimport/screen/model/SimContactUiModel.kt
```diff
@@ -0,0 +1,21 @@
+package com.android.contacts.ui.simimport.screen.model
+
+import androidx.compose.runtime.Immutable
+import kotlinx.collections.immutable.ImmutableList
+
+@Immutable
+data class SimContactUiModel(
```
m4pl: public while neighbours are `internal`. Same for both test factories.

src/com/android/contacts/ui/simimport/common/SimImportAccountPicker.kt
```diff
@@ -0,0 +1,179 @@
...
+            .size(24.dp),
+    )
+}
+
+@PreviewLightDark
+@Composable
+private fun SimImportAccountPickerSinglePreview() {
+    val resources = LocalResources.current
+    val context = LocalContext.current
+    val icon = ResourcesCompat.getDrawable(
```
m4pl: `icon` is unused.

src/com/android/contacts/domain/sim/model/SimContactsResult.kt
```diff
@@ -0,0 +1,13 @@
...
+import com.android.contacts.domain.accounts.model.AccountModel
+import com.android.contacts.model.SimContact
+import kotlinx.collections.immutable.ImmutableList
+import kotlinx.collections.immutable.ImmutableMap
+import kotlinx.collections.immutable.persistentListOf
+import kotlinx.collections.immutable.persistentMapOf
+
+internal data class SimContactsResult(
+    val contacts: ImmutableList<SimContact> = persistentListOf(),
+    val existingContactsInAccounts: ImmutableMap<AccountModel, Set<SimContact>> = persistentMapOf(),
```
m4pl: Immutable collections are a Compose concern, and both get rebuilt anyway at `SimImportViewModel`.

src/com/android/contacts/ui/simimport/screen/SimImportScreen.kt
```diff
@@ -0,0 +1,364 @@
...
+            Modifier.padding(
+                top = contentPadding.calculateTopPadding(),
+                start = contentPadding.calculateStartPadding(layoutDirection),
+                end = contentPadding.calculateEndPadding(layoutDirection),
+            ),
+        ) {
+            uiState.currentAccount?.let {
+                SimImportAccountPicker(
+                    list = uiState.accounts ?: persistentListOf(),
+                    current = uiState.currentAccount,
```
m4pl, line 114: Could this be `current = it`? Right now `uiState.currentAccount` is passed instead, and since a safe call doesn't smart cast, the param has to stay nullable with a `?: return` at `SimImportAccountPicker.kt`. Using `it` lets both go.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,302 @@
...
+    private val _effects = MutableSharedFlow<Effect>(extraBufferCapacity = 1)
+    override val effects: Flow<Effect> = _effects.asSharedFlow()
+
+    private val _uiState = MutableStateFlow(State())
+    override val uiState = _uiState.asStateFlow()
+
+    private val simContacts =
+        MutableStateFlow<ImmutableList<SimContactUiModel>?>(null)
+    private val existingContacts =
+        MutableStateFlow<ImmutableMap<AccountModel, ImmutableSet<Int>>>(persistentMapOf())
```
m4pl: `PersistentMap` here would drop the `toPersistentMap()`.

src/com/android/contacts/di/sim/SimImportBindsModule.kt
```diff
@@ -0,0 +1,20 @@
+package com.android.contacts.di.sim
```
m4pl: Binds a UI mapper from `di/sim`. Maybe `di/simimport/`?

app/src/test/kotlin/com/android/contacts/tests/AccountModelFactory.kt
```diff
@@ -0,0 +1,16 @@
+package com.android.contacts.tests
+
+import com.android.contacts.domain.accounts.model.AccountModel
+import kotlin.random.Random
+
+internal object AccountModelFactory {
+    fun build(
+        name: String = "Account ${Random.nextInt().toString().take(4)}",
```
m4pl: `take(4)` on a signed int leaves only ~9900 names, so two accounts can come out identical and some tests need them to differ. Flaky. A counter instead? Same at `AccountDisplayModelFactory` and `AccountUiModelFactory`.

src/com/android/contacts/ui/simimport/screen/SimImportScreen.kt
```diff
@@ -0,0 +1,364 @@
...
+internal fun SimImportScreen(
+    effectHandler: SimImportEffectHandler,
+    modifier: Modifier = Modifier,
+    screenModel: SimImportScreenModel = viewModel<SimImportViewModel>(),
+) {
+    val uiState by screenModel.uiState.collectAsStateWithLifecycle()
+    val effectHandler by rememberUpdatedState(effectHandler)
+
+    LaunchedEffect(screenModel) {
+        screenModel.effects.collect(effectHandler::handle)
```
m4pl, line 76: `effectHandler::handle` is a bound reference, so `rememberUpdatedState` has no effect.

src/com/android/contacts/domain/sim/usecase/LoadSimContacts.kt
```diff
@@ -0,0 +1,62 @@
...
+            .map { load(subscriptionId) }
+            .catch {
+                if (it is CancellationException) {
```
m4pl, line 34: This swallows the exception instead of rethrowing it. Throw it, or drop the branch?

src/com/android/contacts/domain/sim/usecase/StartSimImport.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.android.contacts.SimImportService
+import com.android.contacts.domain.accounts.model.AccountModel
+import com.android.contacts.model.SimContact
+import com.android.contacts.model.account.AccountWithDataSet
+import dagger.hilt.android.qualifiers.ApplicationContext
+import javax.inject.Inject
+import kotlinx.collections.immutable.ImmutableList
+
+internal fun interface StartSimImport {
+    operator fun invoke(
```
m4pl, line 12: `operator` is here but not in `LoadSimContactsImpl` / `BuildBroadcastReceiverFlowImpl`.

src/com/android/contacts/ui/core/Preview.kt
```diff
@@ -0,0 +1,43 @@
...
+}
+
+@Composable
+internal fun ContactsPreviewColumn(
+    modifier: Modifier = Modifier,
+    content: @Composable () -> Unit,
+) {
+    ContactsPreviewTheme(modifier = modifier) {
+        Column(
+            verticalArrangement = Arrangement.spacedBy(16.dp),
```
m4pl: Real list uses `spacedBy(1.dp)`, so cell previews don't match.

src/com/android/contacts/domain/accounts/model/AccountDisplayModel.kt
sdsantos: It comes as a Drawable from inside the AccounType. Usually it starts from an icon resource, but sometimes it gets mutated before reaching here.

Maybe we can agree to mark the `AccountUiModel` as `@Stable`, knowning it's unlikely it's actually going to change, without the type changing as well?

The alternative is ignoring the icon we get from the AccountType, and attributing our own icon resources for every known AccountType (external accounts are the most complex scenario).

.editorconfig
sdsantos: Copied from https://github.com/GrapheneOS/app-docs/pull/2. Already fixed here and left a comment there.

.editorconfig
sdsantos: Copied from https://github.com/GrapheneOS/app-docs/pull/2. Already fixed here and left a comment there.

src/com/android/contacts/di/sim/SimImportBindsModule.kt
sdsantos: Done, but it's tough to distinguish domain/ui binds where they all share similar package names.

src/com/android/contacts/domain/sim/usecase/StartSimImport.kt
sdsantos: But `operator` is present in `LoadSimContactsImpl` and `BuildBroadcastReceiverFlowImpl` 🤔 
Or am I missing something?

app/src/test/kotlin/com/android/contacts/tests/AccountModelFactory.kt
sdsantos: You're right. And I feel it's a bad idea to rely on the Factory for uniqueness. Better leave that responsibility to the test, if it really needs multiple unique objects.

src/com/android/contacts/ui/simimport/screen/model/AccountUiModel.kt
sdsantos: I can do a mapper. I just felt it was a simple enough 1-1 mapping between 2 models already written in Kotlin.

The goal was to avoid adding the `@Immutable` annotation to `AccountModel` since it's, in theory, a domain model and not a UI model.

src/com/android/contacts/domain/sim/usecase/LoadSimContacts.kt
sdsantos: I chose to stop there since the PR was already getting pretty long (and it was my first). So I didn't touch the `data` layer. But the next time someone touches this code we should start working on the `data` layer and introduce some more abstractions.

src/com/android/contacts/domain/sim/usecase/StartSimImport.kt
m4pl: My mistake. Ignore this one.

src/com/android/contacts/ui/simimport/screen/model/AccountUiModel.kt
m4pl: Could `val account: AccountModel` replace the three private fields and the getter?

app/src/test/kotlin/com/android/contacts/tests/AccountModelFactory.kt
m4pl: `SimContactFactory.kt:8` and `SimContactUiModelFactory.kt:10` still default `recordNumber` to `Random.nextInt()`.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
m4pl: `loadAccounts()` and `loadSimContacts()` are still collected in init with `launchIn(viewModelScope)`, so the receiver stays registered for the ViewModel's lifetime and `WhileSubscribed` only gates the combine over the `MutableStateFlow`s.

src/com/android/contacts/domain/accounts/model/AccountDisplayModel.kt
m4pl: `getDisplayIcon()` returns a new `Drawable` on every call, so `accounts.contains(it)` in `findCurrentAccount` never matches. `@Stable` wouldn't change that.

Own icon set isn't needed though. `AccountInfo.getType()` gives you the `AccountType`, where `syncAdapterPackageName` and `iconRes` are public. Those plus the grey tint from `FallbackAccountType` / `SimAccountType` are enough for the UI to build it.

src/com/android/contacts/ui/simimport/screen/SimImportScreen.kt
```diff
@@ -0,0 +1,375 @@
...
+        modifier = modifier,
+    ) { contentPadding ->
+        Column(
+            Modifier.padding(
+                top = contentPadding.calculateTopPadding(),
+                start = contentPadding.calculateStartPadding(layoutDirection),
+                end = contentPadding.calculateEndPadding(layoutDirection),
+            ),
+        ) {
+            if (uiState is State.Ready) {
```
m4pl: Should the picker still show on the "no contacts" screen? It's only rendered for `State.Ready` now.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,313 @@
...
+                    isSelected -> currentSelectedContacts + contact.recordNumber
+                    else -> currentSelectedContacts - contact.recordNumber
+                }
+                )
+        }
+    }
+
+    private fun selectAll() {
+        val account = currentAccount.value ?: return
+        val state = (uiState.value as? State.Ready) ?: return
```
m4pl: `selectAll()` and `getSelectedContacts()` read `uiState.value`, so the handlers depend on the state they produce. Could they use `simContacts` + `existingContacts` + `currentAccount` directly?

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,313 @@
...
+                }.toImmutableList(),
+            contactsAlreadyImported = contactsAlreadyImported
+                .map(simContactUiModelMapper::map)
+                .toImmutableList(),
+        )
+    }
+
+    private fun restoreSelectedContacts(): Map<AccountModel, Set<Int>> {
+        val entries = savedStateHandle.get<List<AccountContactsEntry>>(KEY_SELECTED_CONTACTS)
+        return entries
```
m4pl: Returns `Map<AccountModel, Set<Int>>` but still builds it with `toImmutableSet()` / `toImmutableMap()` / `persistentMapOf()`.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,313 @@
...
+            is Action.AccountChanged -> changeAccount(action.account)
+            is Action.ContactSelectionChanged ->
+                changeContactSelection(action.contact, action.isSelected)
+            Action.SelectAllClicked -> selectAll()
+            Action.DeselectAllClicked -> deselectAll()
+            Action.ImportClicked -> startImport()
+        }
+    }
+
+    @Suppress("detekt:ReturnCount")
```
m4pl: Would a `when` avoid the suppression here?

m4pl (commented):
Added a few comments in the existing threads. Also worth another pass on unused imports: `kotlin.random.Random` in `AccountModelFactory.kt:4`, `AccountDisplayModelFactory.kt:6`, `AccountUiModelFactory.kt:6`, `AccountModel` in `AccountDisplayModelFactory.kt:5`, `ImmutableList` in `StartSimImport.kt:10`. ktlint doesn't flag them here.

src/com/android/contacts/ui/simimport/screen/SimImportScreen.kt
sdsantos: We could show it yeah, we have the info.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
sdsantos: I've split the method in two steps.

src/com/android/contacts/domain/accounts/model/AccountDisplayModel.kt
sdsantos: I was afraid of going that deep, but I should have tackled the whole things, it's not that big the amount of fields we need to pass along.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
sdsantos: Me and @m4pl and decided to keep the flows as is, to avoid increasing the complexity of screen that's opened few times, and for a few seconds.

m4pl (approved):
Looks good. Thanks!

src/com/android/contacts/ui/common/components/AccountIcon.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.compose.ui.unit.dp
+import com.android.contacts.R
+import com.android.contacts.domain.accounts.model.AccountIconData
+import com.android.contacts.domain.accounts.model.AccountModel
+import com.android.contacts.ui.core.ContactsPreviewColumn
+import com.android.contacts.ui.simimport.screen.model.AccountUiModel
+import com.google.accompanist.drawablepainter.rememberDrawablePainter
+
+@Composable
+internal fun AccountIcon(account: AccountUiModel) {
```
RankoR, line 20: `modifier: Modifier = Modifier` is missing, please check other places too.

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,308 @@
...
+            }.toImmutableList(),
+            contactsAlreadyImported = contactsAlreadyImported
+                .map(simContactUiModelMapper::map)
+                .toImmutableList(),
+        )
+    }
+
+    private fun restoreDeselectedContacts(): Map<AccountModel, Set<Int>> {
+        val entries = savedStateHandle.get<List<AccountContactsEntry>>(KEY_DESELECTED_CONTACTS)
+        return entries?.associate { it.account to it.contactNumbers.toImmutableSet() }.orEmpty()
```
RankoR, line 206: Return type should be changed or `.toImmutableSet()` dropped

src/com/android/contacts/ui/simimport/screen/SimImportViewModel.kt
```diff
@@ -0,0 +1,308 @@
...
+
+    private val subscriptionId: Int = savedStateHandle[EXTRA_SUBSCRIPTION_ID] ?: run {
+        Log.w(TAG, "$EXTRA_SUBSCRIPTION_ID missing, defaulting to ${SimCard.NO_SUBSCRIPTION_ID}")
+        SimCard.NO_SUBSCRIPTION_ID
+    }
+
+    init {
+        if (subscriptionId != SimCard.NO_SUBSCRIPTION_ID) {
+            loadSimCards()
+                .filter { simCards -> simCards.none { it.subscriptionId == subscriptionId } }
```
RankoR, line 98: `LoadSimCardsImpl.load()` returns `emptyList()` both on `UnsupportedOperationException` and whenever `SimContactDaoImpl.canReadSimContacts()` is false - which includes SIM state != SIM_STATE_READY, i.e. airplane mode. `none` on an empty list is true, and `LoadSimCards` has its own `onStart { emit(Unit) }`, so this evaluates immediately  on open. A failed or unavailable read is indistinguishable from a removed SIM: screen closes with RESULT_CANCELED and logs "SIM card removed". 

Can be fixed like this:

```
.filter { simCards -> simCards.isNotEmpty() && simCards.none { it.subscriptionId == subscriptionId } }
```

src/com/android/contacts/domain/accounts/model/AccountIconData.kt
```diff
@@ -0,0 +1,11 @@
+package com.android.contacts.domain.accounts.model
+
+import androidx.annotation.StringRes
+import com.android.contacts.R
+
+internal data class AccountIconData(
+    @StringRes val titleRes: Int = R.string.account_phone,
+    val iconRes: Int,
```
RankoR, line 8: `@DrawableRes` missed

RankoR (changes requested):

RankoR (approved):

PR #36, opened 2026-07-23
Rewrite ImportDialog with Kotlin + Compose

sdsantos:
Related with #35 

|Before|After|
|-|-|
|Screenshot_20260722_142642|Screenshot_20260723_123735|


src/com/android/contacts/ui/interactions/importing/screen/ImportDialog.kt
```diff
@@ -0,0 +1,346 @@
...
+@Composable
+internal fun ImportDialogContent(
+    uiState: State,
+    onAction: (Action) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    @OptIn(ExperimentalMaterial3Api::class)
+    ModalBottomSheet(
+        onDismissRequest = { onAction(Action.Dismiss) },
+        sheetState = rememberStandardBottomSheetState(
```
RankoR: This should be `rememberModalBottomSheetState(skipPartiallyExpanded = true)`, otherwise "back" gesture can partially collapse the sheet instead of closing it

src/com/android/contacts/ui/interactions/importing/screen/ImportEffectHandler.kt
```diff
@@ -0,0 +1,56 @@
...
+    override fun handle(effect: Effect) {
+        when (effect) {
+            Effect.Close -> {
+                activity.finish()
+            }
+
+            is Effect.OpenSimImport -> {
+                activity.startActivity(
+                    UIIntents.getSimImportIntent(activity, effect.subscriptionId),
+                )
```
RankoR, line 28: Probably need to add `finish` here

src/com/android/contacts/ui/interactions/importing/screen/mapper/SimCardOptionMapper.kt
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    private fun getFormattedPhone(simCard: SimCard): String? {
+        return (simCard.getFormattedPhone() ?: simCard.phone)
+            ?.ifEmpty { null }
+            ?.let { phone ->
+                bidiFormatter.unicodeWrap(
+                    PhoneNumberUtilsCompat.createTtsSpannable(phone),
+                    TextDirectionHeuristicsCompat.LTR,
+                )?.toString()
```
RankoR: `toString` removes TTS spans. Would be better to render it as `AnnotatedString` with `VerbatimTtsAnnotation`.

src/com/android/contacts/ui/interactions/importing/screen/ImportDialog.kt
```diff
@@ -0,0 +1,346 @@
...
+            Icon(
+                painter = painterResource(R.drawable.quantum_ic_sim_card_vd_theme_24),
+                contentDescription = null,
+                modifier = Modifier.padding(end = 12.dp),
+            )
+            Column {
+                Text(
+                    text = stringResource(
+                        R.string.import_from_sim_summary_fmt,
+                        option.name ?: (index + 1),
```
RankoR: Old code had a special case for one sim card, it used `R.string.import_from_sim` there. Without it, result may look weird, for example, `SIM SIM 1`, `SIM Verizon`, etc

RankoR (changes requested):

src/com/android/contacts/ui/interactions/importing/screen/ImportDialog.kt
sdsantos: I actually done that in the next PRs and forgot to backport it here. Now it's actually: `confirmValueChange = { it != SheetValue.PartiallyExpanded },`

RankoR: It should be `rememberModalBottomSheetState(skipPartiallyExpanded = true)`. If window exceeds half the screen, it will be undismissable by "back" otherwise.

src/com/android/contacts/ui/interactions/importing/screen/mapper/SimCardOptionMapper.kt
```diff
@@ -0,0 +1,33 @@
...
+        return SimCardOption(
+            subscriptionId = simCard.subscriptionId,
+            name = simCard.displayName?.toString(),
+            contactsCount = simCard.contacts?.size,
+            phone = annotatedPhone(simCard),
+        )
+    }
+
+    private fun annotatedPhone(simCard: SimCard): AnnotatedString {
+        val phone = simCard.getFormattedPhone() ?: simCard.phone
```
RankoR: `simCard.phone` - potential NPE below if `simCard.phone` is `null`.

src/com/android/contacts/ui/interactions/importing/screen/ImportDialog.kt
```diff
@@ -0,0 +1,367 @@
...
+
+@Composable
+private fun simCardOptionDescription(option: SimCardOption): AnnotatedString? {
+    fun CharSequence.toAnnotatedString(): AnnotatedString {
+        return buildAnnotatedString { append(this@toAnnotatedString) }
+    }
+
+    return when {
+        option.contactsCount != null && !option.phone.isNullOrBlank() -> {
+            expandStringTemplate(
```
RankoR: TTS annotation will be still destroyed. `expandStringTemplate` -> `TextUtils.expandTemplate` -> `SpannableStringBuilder.replace`, it copies spans only from `Spanned` inputs, and AnnotatedString is not `Spanned`.

RankoR (changes requested):

src/com/android/contacts/ui/interactions/importing/screen/ImportDialog.kt
sdsantos: Sorry, I missed that `rememberStandardBottomSheetState` was different from `rememberModalBottomSheetState` 🤦

RankoR (approved):

res/values/styles.xml
```diff
@@ -20,6 +20,13 @@
         <item name="alertDialogTheme">@style/ContactsAlertDialogThemeAppCompat</item>
     </style>
 
+    <style name="Theme.Compose.Dialog" parent="Theme.Compose">
+        <item name="android:windowBackground">@android:color/transparent</item>
+        <item name="android:windowContentOverlay">@null</item>
```
m4pl, line 25: `Theme.Material.Light` already sets `windowContentOverlay` to `@null` in the platform themes, so this line is a no-op. Optional.

src/com/android/contacts/ui/interactions/importing/screen/ImportDialog.kt
```diff
@@ -0,0 +1,364 @@
...
+    accountChosen: AccountModel? = null,
+) {
+    val uiState by screenModel.uiState.collectAsStateWithLifecycle()
+
+    LaunchedEffect(screenModel) {
+        screenModel.effects.collect(effectHandler::handle)
+    }
+
+    LaunchedEffect(screenModel, accountChosen) {
+        accountChosen?.let { screenModel.onAction(Action.AccountChosen(accountChosen)) }
```
m4pl, line 72: `it` is unused here, the body reads `accountChosen` again. `?.let { screenModel.onAction(Action.AccountChosen(it)) }`?

src/com/android/contacts/ui/interactions/importing/screen/ImportDialog.kt
```diff
@@ -0,0 +1,364 @@
...
+) {
+    OptionCell(
+        isFirst = true,
+        isLast = uiState.simCardOptions.isNullOrEmpty(),
+        onClick = { onAction(Action.VCardClick) },
+        modifier = Modifier.testTag(IMPORT_VCARD_BUTTON_TEST_TAG),
+    ) {
+        Row(verticalAlignment = Alignment.CenterVertically) {
+            Icon(
+                painter = painterResource(R.drawable.ic_upload_file_24),
```
m4pl, line 187: `Icons.Outlined.UploadFile` and `Icons.Outlined.SimCard` are already available. That drops the new drawable and the quantum asset at line `215`.

src/com/android/contacts/ui/interactions/importing/screen/ImportDialog.kt
```diff
@@ -0,0 +1,364 @@
...
+
+    if (parts.isEmpty()) {
+        return null
+    }
+
+    return buildAnnotatedString {
+        parts.forEachIndexed { index, part ->
+            append(part)
+            if (index != parts.lastIndex) {
+                append(" • ")
```
m4pl, line 296: `00f3ca8` removed `import_from_sim_secondary_template`, which held the bullet and the ordering for every locale. Keep it and format through `messageFormatResource` instead of hardcoding " • "?

src/com/android/contacts/di/core/CoreProvidesModule.kt
```diff
@@ -117,4 +118,10 @@ internal class CoreProvidesModule {
     ): SubscriptionManager {
         return context.getSystemService(SubscriptionManager::class.java)
     }
+
+    @Provides
+    @Reusable
+    fun provideBidiFormatter(): BidiFormatter {
+        return BidiFormatter.getInstance()
```
m4pl, line 125: `getInstance()` freezes the locale direction at creation, so a `@Reusable` singleton keeps the app-start direction after a language change.

src/com/android/contacts/domain/sim/usecase/LoadSimCards.kt
```diff
@@ -43,6 +41,7 @@ internal class LoadSimCardsImpl @Inject constructor(
                 coroutineDispatcher.asExecutor(),
                 listener,
             )
+            trySend(Unit)
```
m4pl, line 44: `addOnSubscriptionsChangedListener` already fires once on registration, so this doubles the initial SIM read. Drop it, or keep it with a comment and `distinctUntilChanged()`?

src/com/android/contacts/ui/interactions/importing/screen/ImportDialog.kt
```diff
@@ -0,0 +1,364 @@
...
+@Composable
+private fun SimCardCell(
+    uiState: State,
+    onAction: (Action) -> Unit,
+    index: Int,
+    option: SimCardOption,
+    isSingleOption: Boolean,
+) {
+    OptionCell(
+        isFirst = index == 0 && uiState.isVCardImportAvailable != true,
```
m4pl, line 208: Position rules are split between `isFirst` here and `isLast` at line `181`, so a third row type means editing both. Pass `isFirst`, `isLast` and `onClick` instead of the whole `uiState`?

src/com/android/contacts/ui/interactions/importing/screen/ImportDialog.kt
```diff
@@ -0,0 +1,364 @@
...
+        isLast = index == uiState.simCardOptions?.lastIndex,
+        onClick = { onAction(Action.SimOptionClick(option)) },
+        modifier = Modifier.testTag(IMPORT_SIM_CARD_BUTTON_TEST_TAG),
```
m4pl, line 211: Same tag on every SIM row. The `LazyColumn` key at line `161` already uses `subscriptionId`.

src/com/android/contacts/ui/interactions/importing/screen/model/ImportUiState.kt
```diff
@@ -0,0 +1,15 @@
+package com.android.contacts.ui.interactions.importing.screen.model
+
+import androidx.compose.runtime.Immutable
+import kotlinx.collections.immutable.ImmutableList
+
+@Immutable
+internal data class ImportUiState(
+    val isVCardImportAvailable: Boolean? = null,
```
m4pl, line 8: Never `null`, so `isLoading` is just `simCardOptions == null`. Not blocking.

app/src/test/kotlin/com/android/contacts/ui/simimport/SimImportViewModelTest.kt
m4pl, line 1: This and `ImportDialogTest.kt` moved into `screen/` in `00f3ca8` but kept the old package. Same for `SimCardOptionMapperTest`.

src/com/android/contacts/domain/sim/usecase/LoadSimCards.kt
m4pl, line 51: The rewrite dropped `isSimImportAllowedForCurrentUser()` from `ImportDialogFragment` and nothing else gates it, so guests can import SIM contacts now. Restore it here so `SimImportActivity` is covered too, with a test?

src/com/android/contacts/ui/interactions/importing/screen/mapper/SimCardOptionMapper.kt
```diff
@@ -0,0 +1,47 @@
...
+    private fun annotatedPhone(simCard: SimCard): AnnotatedString? {
+        val phone = (simCard.getFormattedPhone() ?: simCard.phone)
+            ?.takeIf { it.isNotBlank() }
+            ?: return null
+        val wrappedPhone = bidiFormatter.unicodeWrap(
+            PhoneNumberUtils.createTtsSpannable(phone),
+            TextDirectionHeuristicsCompat.LTR,
+        )
+            .takeIf { it.isNotBlank() }
+            ?: phone
```
m4pl, line 39: phone is non blank here and `unicodeWrap` only returns `null` for `null` input, so this branch is unreachable. Optional.

src/com/android/contacts/ui/interactions/importing/screen/ImportEffectHandler.kt
```diff
@@ -0,0 +1,57 @@
...
+
+            is Effect.OpenVCardImport -> {
+                AccountSelectionUtil.doImportFromVcfFile(
+                    activity,
+                    effect.account?.let { account ->
+                        AccountWithDataSet(
+                            account.name,
+                            account.type,
+                            account.dataSet,
+                        )
```
m4pl, line 50: `AccountModel` to `AccountWithDataSet` is built by hand here and again at `StartSimImport.kt:32`, while the opposite direction has `AccountModelMapper`. A mapper for this direction would keep both in step.

src/com/android/contacts/ui/interactions/importing/screen/ImportEffectHandler.kt
```diff
@@ -0,0 +1,57 @@
...
+            }
+
+            Effect.OpenSelectAccount -> {
+                SelectAccountDialogFragment.show(
+                    @Suppress("DEPRECATION")
+                    activity.fragmentManager,
+                    R.string.dialog_new_contact_account,
+                    AccountTypeManager.AccountFilter.CONTACTS_INSERTABLE,
+                    null,
+                )
```
m4pl, line 39: The `android.app.DialogFragment` base class of `SelectAccountDialogFragment` is the only reason for this `@Suppress`. Both callers are `AppCompatActivity` already, so switching it to `androidx.fragment.app.DialogFragment` with `supportFragmentManager` is contained. Up to you whether it belongs here.

src/com/android/contacts/ui/interactions/importing/screen/ImportViewModel.kt
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    private fun emitEffect(effect: Effect) {
+        _effects.tryEmit(effect)
+    }
+
+    private fun onVCardClick() {
+        // Accounts should be loaded by now
+        val accounts = accounts ?: return
```
m4pl, line 92: `isLoading` only covers `simCardOptions`, so the `.vcf` row can appear before `loadAccounts` has emitted, and a tap in that window returns here and does nothing. Accounts in the state, like `SimImportViewModel.kt:68`, would close it. Not covered by tests.

src/com/android/contacts/domain/vcard/usecase/CanImportFromVCard.kt
```diff
@@ -0,0 +1,18 @@
...
+internal fun interface CanImportFromVCard {
+    operator fun invoke(): Boolean
+}
+
+internal class CanImportFromVCardImpl @Inject constructor(
+    @param:ApplicationContext private val context: Context,
+) : CanImportFromVCard {
+    override fun invoke(): Boolean {
+        return context.resources.getBoolean(R.bool.config_allow_import_from_vcf_file)
+    }
```
m4pl, line 17: This reads one resource bool and is called from the VM constructor at `ImportViewModel.kt:46`, so it runs on the main thread. `SettingsAvailabilityRepository` does the same kind of read in the data layer. Moving it there drops this class and `VCardBindsModule`.

src/com/android/contacts/ui/interactions/importing/screen/mapper/SimCardOptionMapper.kt
```diff
@@ -0,0 +1,47 @@
...
+            phone = annotatedPhone(simCard),
+        )
+    }
+
+    private fun annotatedPhone(simCard: SimCard): AnnotatedString? {
+        val phone = (simCard.getFormattedPhone() ?: simCard.phone)
+            ?.takeIf { it.isNotBlank() }
+            ?: return null
+        val wrappedPhone = bidiFormatter.unicodeWrap(
+            PhoneNumberUtils.createTtsSpannable(phone),
```
m4pl, line 35: `AnnotatedString` cannot hold the Android span this returns, so it is dropped at the append on line `43` and only the `VerbatimTtsAnnotation` below reaches `TalkBack`. The call has no effect on how the number is read. Remove it?

m4pl (changes requested):

src/com/android/contacts/ui/interactions/importing/screen/ImportDialog.kt
sdsantos: This solution came after these discussions (https://github.com/GrapheneOS/platform_packages_apps_Contacts/pull/36#discussion_r3926354809 + https://github.com/GrapheneOS/platform_packages_apps_Contacts/pull/36#discussion_r3959459551) and some options I chatted with @RankoR.

`messageFormatResource` loses our TTS compose annotation, unless we rewrite it use AnnotationString instead of Spanned. Since this was already a sort of duplicated string, and all locales kept the same ordering and bullet divider, I felt this was a simpler solution overall, with nothing lost in practice.

src/com/android/contacts/ui/interactions/importing/screen/ImportEffectHandler.kt
sdsantos: This is already dropped in https://github.com/GrapheneOS/platform_packages_apps_Contacts/pull/38 so I don't think we need to worry about it here.

src/com/android/contacts/ui/interactions/importing/screen/ImportViewModel.kt
sdsantos: Right. I copied the old behaviour that basically assumed the accounts would be loaded by the time the user performed an action, but it's not ideal.

But in https://github.com/GrapheneOS/platform_packages_apps_Contacts/pull/39/, this ViewModel is no longer responsible for loading accounts. An account is always explicitly selected, we no longer skip that step.

m4pl (approved):
`SimImportViewModelTest.kt` and `SimImportScreenTest.kt` still use package `com.android.contacts.ui.simimport`, while `SimImportViewModel` and `SimImportScreen` are in `ui.simimport.screen`. `LoadSimCardsImplTest.kt` also ended up in `domain.sim.usercase` instead of `usecase`. Not blocking.

PR #38, opened 2026-07-28
 Rewrite SelectAccountDialog with Kotlin + Compose

sdsantos:
Based on another branch. Compare diff: https://github.com/sdsantos/platform_packages_apps_Contacts/compare/import-dialog...sdsantos:platform_packages_apps_Contacts:select-account

Related with #35 

For now, you can reach the new dialog through:
- Settings > Import > .vcf file (if you have more than 1 account)
- Home side menu > Create Label (if you have at least 1 account)

|Before|After|
|-|-|
|||


src/com/android/contacts/domain/sim/usecase/LoadSimCards.kt
```diff
@@ -11,23 +11,27 @@ import kotlinx.coroutines.asExecutor
...
     private val simContactDao: SimContactDao,
     @param:SimReadDispatcher private val coroutineDispatcher: CoroutineDispatcher,
 ) : LoadSimCards {
 
     override operator fun invoke(): Flow<List<SimCard>> {
+        if (!canUserImportFromSim()) {
+            return flowOf(emptyList())
+        }
+
         return onSubcriptionsChange()
```
RankoR, line 34: `onStart` was useful here, because it's possible that flow never emits - in this case user will see endless spinner

src/com/android/contacts/ui/interactions/account/screen/SelectAccountDialog.kt
```diff
@@ -0,0 +1,212 @@
...
+                        }
+                    }
+                }
+            }
+        },
+    )
+}
+
+@Composable
+private fun ImportOptionsList(
```
RankoR, line 115: I guess it should be `AccountsList` or something like it.

src/com/android/contacts/ui/interactions/account/screen/SelectAccountDialog.kt
```diff
@@ -0,0 +1,212 @@
...
+    uiState: State,
+    onAction: (Action) -> Unit,
+) {
+    LazyColumn(
+        verticalArrangement = Arrangement.spacedBy(1.dp),
+        modifier = Modifier.fillMaxWidth(),
+    ) {
+        itemsIndexed(
+            items = uiState.accounts.orEmpty(),
+            key = { _, account -> "account_${account.hashCode()}" },
```
RankoR, line 125: `account` itself could be a key

RankoR (changes requested):

src/com/android/contacts/domain/sim/usecase/LoadSimCards.kt
sdsantos: From the `SubscriptionManager.addOnSubscriptionsChangedListener` documentation (first spotted by @m4pl):
> The SubscriptionManager.OnSubscriptionsChangedListener.onSubscriptionsChanged() method will also be invoked once initially when calling this method.

So we shouldn't need the `onStart`.

PR #39, opened 2026-08-04
 Rewrite ImportVCardActivity to Kotlin + Compose 

### LINKED ISSUE #35: Rewrite VCF file import to Kotlin + Compose

sdsantos:
Based on another branch. Compare diff: https://github.com/sdsantos/platform_packages_apps_Contacts/compare/select-account...sdsantos:platform_packages_apps_Contacts:import-vcard

Closes https://github.com/GrapheneOS/platform_packages_apps_Contacts/issues/35

The ImportVCard user flow was simplified:
- Now the ImportVCardActivity is always the responsible for selecting the account
- The behavior no longer changes if it's called from an external app
- The import confirmation dialog was removed, and the confirmation is done by selecting the destination account, which is always shown, even if there's only one destination account

The user flow is now:
- Verifies and requests permissions if not granted
- Selects provided VCard file from another app, or opens the select files system screen
- Selects account
- Parses VCard(s) to verify if they are valid
- Starts background import

https://github.com/user-attachments/assets/0ae429d8-65b3-4711-acc1-eee2c790a8f0


astrovolt:
I'm just a user, but wanted to give a little bit of input if possible on the video displayed here on this PR.
As seen, when the "Import contacts" dialog is opened, a grey screen covers the background. This screen seems to push from the bottom up alongside the dialog which is unusual.

When you look at the rest of the OS interactions, opening dialogs will darken the entire background with a grey/dark grey overlay effect over a short period of time, all at once. So this animation implemented here seems inconsistent.

PR #56, opened 2026-08-08
Settings Screen: Java/XML to Kotlin/Compose migration

### LINKED ISSUE #41: Rewrite settings screen to Compose

m4pl:
Closes #41 

### Screenshots
| Before | After |
|--------|-------|
|Screenshot_20260808_150454|Screenshot_20260808_150312|

### Video
https://github.com/user-attachments/assets/c317fb64-f701-4a98-b465-83947d465ea6


src/com/android/contacts/ui/settings/about/ui/AboutScreen.kt
```diff
@@ -0,0 +1,95 @@
+package com.android.contacts.ui.settings.about.ui
```
sdsantos: Not sure this needs the extra `ui` package.

src/com/android/contacts/ui/settings/about/AboutScreen.kt
```diff
@@ -0,0 +1,95 @@
...
+                            contentPadding.calculateStartPadding(layoutDirection),
+                        end = SETTINGS_HORIZONTAL_PADDING +
+                            contentPadding.calculateEndPadding(layoutDirection),
+                        top = SETTINGS_TOP_PADDING + contentPadding.calculateTopPadding(),
+                        bottom = contentPadding.calculateBottomPadding(),
+                    ),
+                ),
+        ) {
+            SettingsCell(
+                title = stringResource(R.string.about_build_version),
```
sdsantos, line 79: It's a new requirement, but I always feel build / version numbers should always be at least selectable, if not copy-able.

src/com/android/contacts/ui/settings/screen/model/SettingsEffect.kt
```diff
@@ -0,0 +1,26 @@
+package com.android.contacts.ui.settings.screen.model
+
+internal sealed interface SettingsEffect {
+
+    sealed interface Host : SettingsEffect
+    sealed interface Message : SettingsEffect
```
sdsantos, line 6: Nice approach, hadn't seen it before 👍

src/com/android/contacts/ui/settings/screen/SettingsContent.kt
```diff
@@ -0,0 +1,83 @@
...
+) {
+    AnimatedContent(
+        targetState = route,
+        transitionSpec = { routeTransition() },
+        label = "settings_navigation",
+        modifier = modifier.background(MaterialTheme.colorScheme.background),
+    ) { currentRoute ->
+        when (currentRoute) {
+            SettingsNavRoute.Main -> {
+                SettingsMainScreen(
```
sdsantos: Naming is definitely hard. 

SettingsScreen > SettingsContent > SettingsMainScreen feels a bit weird to me, but do we have an alternative name for this parent Screens than hold navigation inside for other screens? Or maybe we call the inner screens Panes like for adaptive layouts?

src/com/android/contacts/ui/settings/screen/SettingsViewModel.kt
```diff
@@ -0,0 +1,184 @@
...
+            else -> Effect.OpenProfile(contactId)
+        }
+
+        emitEffect(effect)
+    }
+
+    private fun selectSortOrder(sortOrder: SortOrder) {
+        viewModelScope.launch {
+            displaySettingsRepository.setSortOrder(sortOrder)
+            refreshState()
```
sdsantos: This we're writing the Repository ourselves here, we could also make them reactive instead: return a flow, and the flow automatically updates on a `set`.

sdsantos (approved):
UI things I found:

1. I never see the snackbar messages after an Import finishes. Is it only broadcasted in special circumstances?

2. I feel the Settings cells should be lighter and not darker than the background. It helps having a greater contrast with the cell description (onSurfaceVariant). Here's an example from the latest Google Messaging app:

|Messaging|New Contacts|
|-|-|
|||

Their space between cells is also slightly lower (1 vs 2 dp), which I find pleasant. And also the top padding is slightly lower (8 vs 16 dp). But I'm not 100% sure those are improvements to be honest.

M00NJ:
Sorry for barging in here. I'd just like to comment on this, since I noticed similar things while making mockups for #37.

> I feel the Settings cells should be lighter and not darker than the background. It helps having a greater contrast with the cell description (onSurfaceVariant). Here's an example from the latest Google Messaging app:

The way it is now is in accordance to the MD3E guidelines. Google themself don't always stick to them however and their implementation is inconsistent across their apps. I'd suggest going with the Material Design guidelines over Google's reference when in question. Apart from that I would argue, that there's more separation and emphasis on each settings item here, giving the screen more structure, compared to Google's implementation. Contrast is still sufficient if you ask me.

> Their space between cells is also slightly lower (1 vs 2 dp), which I find pleasant. 

I think this might be an optical illusion as a result of the color roles being reversed here.

sdsantos:
> Sorry for barging in here. I'd just like to comment on this, since I noticed similar things while making mockups for #37.
> 
> > I feel the Settings cells should be lighter and not darker than the background. It helps having a greater contrast with the cell description (onSurfaceVariant). Here's an example from the latest Google Messaging app:
> 
> The way it is now is in accordance to the MD3E guidelines. Google themself don't always stick to them however and their implementation is inconsistent across their apps. I'd suggest going with the Material Design guidelines over Google's reference when in question. Apart from that I would argue, that there's more separation and emphasis on each settings item here, giving the screen more structure, compared to Google's implementation. Contrast is still sufficient if you ask me.

I feel even the MD3E guidelines has examples with both approaches, that's why I started looking at concrete apps. But you are right that in general the Color guideline is Background (outside) > SurfaceContainer (inside). Maybe I should switch here as well then (https://github.com/GrapheneOS/platform_packages_apps_Contacts/pull/33), since I was matching the Google Contacts app.

But maybe we can tone it down a bit? We're defaulting to `surfaceContainerHigh` on the cells, when a `surfaceContainer` would do just fine. I feel we just need the Lows and Highs when we have more hierarchies?

I tested the text color contrast with the accessibility scanner and `onSurfaceVariant` passes, so I guess although there's less contrast, it's enough as you said.

> > Their space between cells is also slightly lower (1 vs 2 dp), which I find pleasant.
> 
> I think this might be an optical illusion as a result of the color roles being reversed here.

Could be yeah. In the guidelines I only find a mention to a `List segment gap` of `2dp`.

sdsantos:
I also noticed the scroll position it's not kept when returning from AboutScreen to the MainSettingsScreen. Not a big deal, but I sort of expected it.

M00NJ:
> I feel even the MD3E guidelines has examples with both approaches, that's why I started looking at concrete apps.

Yes, it seems not completely clear cut. I also looked around for more examples. The new Gallery app for example matches the current state. Interestingly the systems settings match the current behavior in dark mode, but in light mode it's reversed and matches Google Messagings implementation.

> But maybe we can tone it down a bit? We're defaulting to `surfaceContainerHigh` on the cells, when a `surfaceContainer` would do just fine. I feel we just need the Lows and Highs when we have more hierarchies?

I wouldn't recommend that. There's quite a bit of thought in those color roles and it might not matter for this screen, but if you want to apply this evenly across all screens and apps, you might end up in a situation where these color roles where reserved for other components and on screen and now you'd have to change those components as well to retain a hierarchy. So I think it's better to pick between the default behavior, how it is now and Google Messagings implementation. Neither is a bad choice, though I would personally lean towards the current one.

src/com/android/contacts/ui/settings/screen/SettingsViewModel.kt
```diff
@@ -0,0 +1,193 @@
...
+    private val refreshTriggers = Channel<Unit>(Channel.CONFLATED)
+
+    private val settingsData: Flow<SettingsData> = refreshTriggers.receiveAsFlow()
+        .onStart { emit(Unit) }
+        .flatMapLatest { getSettingsData() }
+
+    private val profile: StateFlow<ProfileData?> = profileRepository.observeProfile()
+        .stateIn(
+            scope = viewModelScope,
+            started = SharingStarted.Eagerly,
```
RankoR: Maybe `WhileSubscribed(STATE_STOP_TIMEOUT_MILLIS)`?

src/com/android/contacts/ui/settings/screen/SettingsNavHost.kt
```diff
@@ -0,0 +1,67 @@
...
+                        snackbarHostState = snackbarHostState,
+                    )
+                }
+
+                SettingsNavRoute.About -> {
+                    AboutScreen(
+                        buildVersion = uiState.buildVersion,
+                        onBuildVersionLongClick = { onAction(Action.BuildVersionLongClicked) },
+                        onLicensesClick = { onAction(Action.LicensesClicked) },
+                        onNavigateBack = { route = SettingsNavRoute.Main },
```
RankoR, line 61: No snackbarHostState here

src/com/android/contacts/ui/settings/screen/SettingsScreen.kt
```diff
@@ -0,0 +1,87 @@
...
+import com.android.contacts.ui.settings.screen.model.SettingsEffect as Effect
+import java.util.Locale
+import kotlinx.coroutines.launch
+
+@Composable
+internal fun SettingsScreen(
+    effectHandler: SettingsEffectHandler,
+    onNavigateBack: () -> Unit,
+    modifier: Modifier = Modifier,
+    screenModel: SettingsScreenModel = viewModel<SettingsViewModel>(),
```
RankoR, line 28: Should be `hiltViewModel<SettingsViewModel>()` I guess

RankoR (changes requested):
Also:

- AppCompat dialogs crash under Theme.Compose, worth adding ` <item name="alertDialogTheme">@style/ContactsAlertDialogThemeAppCompat</item>` to theme
- `accompanist-drawablepainter` and `kotlin-parcelize` seem to be dead dependencies now
- `lint-baseline.xml` references deleted files

src/com/android/contacts/ui/settings/screen/SettingsScreen.kt
m4pl: `hiltViewModel()` will be needed when we move to real navigation. For now the owner is the Hilt activity, so `viewModel()` works.

RankoR (approved):

PR #60, opened 2026-08-11
Debug Menu: seed and clear test data

### LINKED ISSUE #59: Debug Menu to generate test contacts
- **Generate test contacts:** Create several hundred contacts with high variability (e.g., some with just a name and phone number, some with/without photos, etc - the data should cover all possible contact fields and their combinations).
- **Delete test data:** Simply remove these synthetic contacts.

sdsantos:
Based on another branch. Compare diff: https://github.com/sdsantos/platform_packages_apps_Contacts/compare/select-account...sdsantos:platform_packages_apps_Contacts:debug-data

Closes https://github.com/GrapheneOS/platform_packages_apps_Contacts/issues/59

Contact photo import is currently failing on debug, so the generated test images aren't visible.

https://github.com/user-attachments/assets/764c9a59-7750-467c-9e23-7620d7e04dd3


res/menu/people_options.xml
```diff
@@ -45,4 +39,10 @@
         android:title="@string/menu_joinAggregate"
         contacts:showAsAction="ifRoom"/>
 
+    <item
+        android:id="@+id/menu_debug"
+        android:title="@string/menu_debug"
+        android:visible="false"
+        contacts:showAsAction="ifRoom"/>
```
m4pl: R.id.menu_debug is not referenced anywhere. Remove?

src/com/android/contacts/activities/PeopleActivity.java
```diff
@@ -205,6 +211,9 @@ public enum ContactsView {
 
     private final Handler mHandler = new Handler();
 
+    @Inject
+    DebugOptionsMenu debugOptionsMenu;
```
m4pl: `Lazy<DebugOptionsMenu>`? Field injection builds the whole graph on every build type.

src/com/android/contacts/drawer/DrawerAdapter.java
```diff
@@ -86,11 +90,15 @@ public class DrawerAdapter extends BaseAdapter {
 
     private List<BaseDrawerItem> mItemsList = new ArrayList<>();
     private AccountDisplayInfoFactory mAccountDisplayFactory;
+    private IsDebugEnabled isDebugEnabled;
 
     public DrawerAdapter(Activity activity) {
         super();
         mInflater = LayoutInflater.from(activity);
         mActivity = activity;
+        isDebugEnabled = EntryPointAccessors
```
m4pl, line 99: The activity is already `@AndroidEntryPoint`, could the flag be passed into the constructor instead?

src/com/android/contacts/ui/debug/DebugOptionsMenu.kt
```diff
@@ -0,0 +1,49 @@
...
+    private fun getOptions(): List<DebugOption> {
+        return DebugOption.entries
+            .filter { it != DebugOption.ExportDatabase || isExportDatabaseAvailable() }
+    }
+
+    @EntryPoint
+    @InstallIn(SingletonComponent::class)
+    interface Provider {
+        fun debugOptionsMenu(): DebugOptionsMenu
+    }
```
m4pl: Provider is unused. Remove?

src/com/android/contacts/domain/debug/usecase/SeedTestData.kt
```diff
@@ -0,0 +1,237 @@
...
+                contentResolver.applyBatch(ContactsContract.AUTHORITY, operations)
+            } catch (e: OperationApplicationException) {
+                Log.w(TAG, "Failed to save test data", e)
+            } catch (e: RemoteException) {
+                Log.w(TAG, "Failed to save test data", e)
+            }
+        }
+    }
+
+    private fun TestContact.toContentValuesList(): List<ContentValues> {
```
m4pl, line 93: No `GroupMembership`, `Event`, `Note`, `Im`, `SipAddress` or `Identity`, and no `TYPE_CUSTOM` label anywhere. Labels and events are a good part of the details screen.

src/com/android/contacts/domain/debug/usecase/SeedTestData.kt
```diff
@@ -0,0 +1,237 @@
...
+                    ContactsContract.Data.MIMETYPE,
+                    ContactsContract.CommonDataKinds.Email.CONTENT_ITEM_TYPE,
+                )
+                put(ContactsContract.CommonDataKinds.Email.ADDRESS, it.value)
+                put(ContactsContract.CommonDataKinds.Email.TYPE, it.type)
+            }
+        }
+    }
+
+    private fun TestContact.postalContentValues(): List<ContentValues> {
```
m4pl: Inserts an empty postal row when city and country are both null.

src/com/android/contacts/domain/debug/usecase/GenerateTestContact.kt
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    private fun randomWebsite(name: String): TestContact.ValueWithType {
+        return TestContact.ValueWithType(
+            "https://${name.lowercase()}.me",
+            WEBSITE_TYPES.random(),
+        )
+    }
+
+    private fun randomPhoto(): ByteArray? {
```
m4pl: Never returns null.

app/src/test/kotlin/com/android/contacts/domain/debug/GenerateTestContactTest.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class GenerateTestContactTest {
+
+    private val subject: GenerateTestContact = GenerateTestContactImpl()
+
+    @Test
+    fun alwaysHaveAtLeastOnePhone() {
```
m4pl, line 18: None of the three can fail: phones is always `(1..3)`, `randomPhone()` hardcodes the prefix, and `NAMES` has no blanks. Injecting `Random` into `GenerateTestContactImpl` would let them seed it and assert real output, including the branches behind `shouldAddExtraProperty()`.

app/src/test/kotlin/com/android/contacts/tests/factory/TestContactFactory.kt
```diff
@@ -0,0 +1,42 @@
...
+
+internal object TestContactFactory {
+    fun build(
+        phones: List<TestContact.ValueWithType> = listOf(
+            TestContact.ValueWithType(
+                value = TestContact.PHONE_PREFIX +
+                    Random.nextInt(999_999).toString().padStart(6, '0'),
+                type = null,
+            ),
+        ),
```
m4pl, line 9: Duplicates `GenerateTestContactImpl.randomPhone()`. Call the generator?

src/com/android/contacts/domain/debug/usecase/SeedTestData.kt
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    private fun saveContacts(account: AccountModel, contacts: List<TestContact>) {
+        contacts.forEach { contact ->
+            val operations = ArrayList<ContentProviderOperation>()
+            operations.add(
+                ContentProviderOperation.newInsert(ContactsContract.RawContacts.CONTENT_URI)
+                    .withValue(ContactsContract.RawContacts.ACCOUNT_NAME, account.name)
+                    .withValue(ContactsContract.RawContacts.ACCOUNT_TYPE, account.type)
+                    .withValue(ContactsContract.RawContacts.DATA_SET, account.type)
```
m4pl: `account.dataSet`, not `account.type`. `SeedTestDataTest.kt:187` misses it because `AccountModelFactory` leaves both null.

src/com/android/contacts/domain/debug/usecase/SeedTestData.kt
```diff
@@ -0,0 +1,237 @@
...
+                contact.toContentValuesList().map { contentValues ->
+                    ContentProviderOperation.newInsert(ContactsContract.Data.CONTENT_URI)
+                        .withValueBackReference(ContactsContract.Data.RAW_CONTACT_ID, 0)
+                        .withValues(contentValues)
+                        .build()
+                },
+            )
+
+            try {
+                contentResolver.applyBatch(ContactsContract.AUTHORITY, operations)
```
m4pl, line 82: `applyBatch` also throws `IllegalArgumentException`.

src/com/android/contacts/ui/debug/DebugOptionsMenu.kt
```diff
@@ -0,0 +1,49 @@
...
+    private val isExportDatabaseAvailable: IsExportDatabaseAvailable,
+    private val exportDatabase: ExportDatabase,
+    private val seedTestData: SeedTestData,
+    private val clearSeededTestData: ClearSeededTestData,
+) {
+    fun show(context: Context) {
+        val optionsLabels = getOptions().map { context.getString(it.titleId) }.toTypedArray()
+        AlertDialog.Builder(context)
+            .setItems(optionsLabels) { _, which ->
+                CoroutineScope(Dispatchers.Default).launch {
```
m4pl: The scope has no exception handler, so the `IllegalArgumentException` from `SeedTestData.kt:81` kills the process. `Dispatchers.Default` is also redundant, both use cases already switch to `IO`.

src/com/android/contacts/domain/debug/usecase/SeedTestData.kt
```diff
@@ -0,0 +1,237 @@
...
+
+    private fun TestContact.relationContentValues(): List<ContentValues>? {
+        return relation?.let {
+            listOf(
+                ContentValues().apply {
+                    put(
+                        ContactsContract.Data.MIMETYPE,
+                        ContactsContract.CommonDataKinds.Relation.CONTENT_ITEM_TYPE,
+                    )
+                    put(ContactsContract.CommonDataKinds.Relation.TYPE, it)
```
m4pl: `Relation.NAME` is never set, so the row renders without a name. See `QuickContactActivity.java:1677`.

src/com/android/contacts/domain/debug/usecase/SeedTestData.kt
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+
+    private fun TestContact.photoContentValues(): List<ContentValues>? {
+        return photo?.let {
+            listOf(
+                ContentValues().apply {
+                    put(
+                        ContactsContract.Data.MIMETYPE,
+                        ContactsContract.CommonDataKinds.Phone.CONTENT_ITEM_TYPE,
```
m4pl: `Photo.CONTENT_ITEM_TYPE`, not Phone.

src/com/android/contacts/domain/debug/usecase/SeedTestData.kt
```diff
@@ -0,0 +1,237 @@
...
+import android.os.RemoteException
+import android.provider.ContactsContract
+import android.util.Log
+import androidx.annotation.VisibleForTesting
+import com.android.contacts.di.core.IoDispatcher
+import com.android.contacts.domain.accounts.model.AccountModel
+import com.android.contacts.domain.accounts.usecase.GetDefaultAccount
+import com.android.contacts.domain.accounts.usecase.LoadAccounts
+import com.android.contacts.domain.debug.model.TestContact
+import com.android.contacts.model.AccountTypeManager.AccountFilter
```
m4pl: `AccountFilter` comes from legacy `AccountTypeManager`, so debug depends on it too. Own enum at the accounts boundary?

src/com/android/contacts/domain/debug/usecase/SeedTestData.kt
```diff
@@ -0,0 +1,237 @@
...
+                },
+            )
+        }
+    }
+
+    companion object {
+        private const val TAG = "SeedTestData"
+
+        @VisibleForTesting
+        var contactsCount = 100
```
m4pl: Mutable static that tests write to at `SeedTestDataTest.kt:59` and `:140`. Constructor param?

src/com/android/contacts/domain/debug/usecase/GenerateTestContact.kt
```diff
@@ -0,0 +1,213 @@
...
+    override fun invoke(): TestContact {
+        val givenName = NAMES.random()
+        val middleName = if (shouldAddExtraProperty()) NAMES.random() else null
+        val familyName = if (shouldAddExtraProperty()) SURNAMES.random() else null
+        val displayName = if (shouldAddExtraProperty()) {
+            listOfNotNull(
+                familyName,
+                middleName,
+                familyName,
+            ).joinToString(" ")
```
m4pl, line 33: `givenName` is missing and `familyName` is used twice. Also gives an empty `DISPLAY_NAME` when both are null.

src/com/android/contacts/domain/debug/usecase/GenerateTestContact.kt
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        return TestContact(
+            phones = (1..3).map { randomPhone() },
+            givenName = givenName,
+            middleName = middleName,
+            familyName = familyName,
+            displayName = displayName,
+            nickname = if (shouldAddExtraProperty()) randomNickname() else null,
+            emails = (0..2).map { randomEmail(givenName) },
```
m4pl: `randomEmail` only depends on the name, so all three emails are the same and `MoreContactUtils.shouldCollapse:62` folds them into one. `(0..2)` here and `(1..3)` on line `35` also always give exactly three.

src/com/android/contacts/domain/debug/model/TestContact.kt
```diff
@@ -0,0 +1,69 @@
...
+        result = 31 * result + (city?.hashCode() ?: 0)
+        result = 31 * result + (country?.hashCode() ?: 0)
+        result = 31 * result + (organization?.hashCode() ?: 0)
+        result = 31 * result + (website?.hashCode() ?: 0)
+        result = 31 * result + (photo?.contentHashCode() ?: 0)
+        return result
+    }
+
+    companion object {
+        const val PHONE_PREFIX = "+15550"
```
m4pl, line 31: The prefix is a contract between the generator and the cleanup, not a property of a contact. Fine as is though.

src/com/android/contacts/domain/debug/model/TestContact.kt
```diff
@@ -0,0 +1,69 @@
...
+    val relation: Int? = null,
+    val website: ValueWithType? = null,
+    val photo: ByteArray? = null,
+) {
+    internal data class ValueWithType(
+        val value: String,
+        val type: Int?,
+    )
+
+    // Due to the ByteArray, we need a custom equals/hashCode
```
m4pl: Wrapping the photo would let the data class generate both, and the suppression could go:
```
internal class Photo(val bytes: ByteArray) {
    override fun equals(other: Any?) = other is Photo && bytes.contentEquals(other.bytes)
    override fun hashCode() = bytes.contentHashCode()
}
```

src/com/android/contacts/domain/debug/usecase/GenerateTestContact.kt
```diff
@@ -0,0 +1,213 @@
...
+            "Magadan", "Majuro", "Mazatlan", "Melbourne", "Mexico City", "Minsk", "Monrovia",
+            "Monterrey", "Moscow", "Muscat", "Nairobi", "New York", "Noumea", "Novosibirsk",
+            "Pago Pago", "Paris", "Perth", "Phoenix", "Port Moresby", "Prague", "Rangoon",
+            "Regina", "Riga", "Riyadh", "Rome", "Santiago", "Sarajevo", "Seoul", "Shanghai",
+            "Singapore", "Skopje", "Sofia", "South Georgia", "St Johns", "Stockholm", "Sydney",
+            "São Miguel", "São Paulo", "São Vicente", "Taipei", "Tallinn", "Tashkent", "Tbilisi",
+            "Tehran", "Tijuana", "Tokyo", "Tongatapu", "Ulaanbaatar", "Urumqi", "Vienna",
+            "Vilnius", "Vladivostok", "Warsaw", "Yakutsk", "Yekaterinburg", "Yerevan", "Zagreb",
+        )
+        private val COUNTRIES = Locale.availableLocales().map { it.displayCountry }.toList()
```
m4pl: `Locale.availableLocales()` includes locales without a region such as `ceb`, `mzn` and `uz-Latn`, and their `displayCountry` is empty. `.filter { it.isNotEmpty() }.distinct()`?

app/src/test/kotlin/com/android/contacts/domain/debug/SeedTestDataTest.kt
```diff
@@ -0,0 +1,200 @@
...
+            "A contact was inserted to the wrong non-default account",
+            operationsList.all { operations ->
+                val firstOperation = operations.first()
+                valuesMatchAccount(firstOperation.values, defaultAccount.account)
+            },
+        )
+    }
+
+    @Test
+    fun usesDefaultAccount_ifThereIsNoDeviceOrDefaultAccount() = runTest {
```
m4pl: Name says default account, the test checks the first one. The assert message on line `120` is copied from the test above.

src/com/android/contacts/domain/debug/usecase/ClearSeededTestData.kt
```diff
@@ -0,0 +1,53 @@
...
+    @param:IoDispatcher private val coroutineDispatcher: CoroutineDispatcher,
+) : ClearSeededTestData {
+    override suspend fun invoke() {
+        withContext(coroutineDispatcher) {
+            val rawContactIds = mutableListOf<Int>()
+
+            contentResolver.query(
+                ContactsContract.Data.CONTENT_URI,
+                arrayOf(ContactsContract.Data.RAW_CONTACT_ID),
+                "${ContactsContract.CommonDataKinds.Phone.NUMBER} LIKE ?",
```
m4pl: `Phone.NUMBER` is `data1`, which every kind reuses, and there is no `MIMETYPE` filter here. A contact whose note or nickname starts with the prefix would be deleted along with all its data. No tests on this class either.

src/com/android/contacts/domain/debug/usecase/SeedTestData.kt
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    /*
+     * Choses the account based on the following order:
+     *  - Device account if available
+     *  - Default account if any is configured
+     *  - First account from the accounts list
+     */
+    private suspend fun getDeviceAccount(): AccountModel? {
+        val accounts = loadAccounts(AccountFilter.ALL).first()
```
m4pl: Crashes with `Cannot add contacts to local or SIM accounts when default account is set to cloud`. `CONTACTS_INSERTABLE` filters those out, the check already lives in `insertableFilter()` at `AccountTypeManager.java:346`.

m4pl (changes requested):

src/com/android/contacts/drawer/DrawerAdapter.java
sdsantos: There's a `DrawerFragment` in between the `PeopleActivity` and the `DrawerAdapter` so I felt it wasn't worth passing it around.

app/src/test/kotlin/com/android/contacts/domain/debug/GenerateTestContactTest.kt
sdsantos: Injecting Random is a good idea. But I felt it wasn't worth it to test this class very thoroughly. I was using this test as a spec, since it's important that all synthetic contacts have a phone, and that is has that prefix, so we can remove them afterwards. The name test is less necessary.

src/com/android/contacts/domain/debug/usecase/SeedTestData.kt
sdsantos: Nice catch.

src/com/android/contacts/domain/debug/usecase/SeedTestData.kt
sdsantos: I trusted the javadoc instead of looking at the code 🤦

app/src/test/kotlin/com/android/contacts/tests/factory/TestContactFactory.kt
sdsantos: I'll continue to avoid using Random in test factories.

src/com/android/contacts/ui/debug/DebugOptionsMenu.kt
sdsantos: Switched to `Dispatchers.Main` to avoid any extra thread switching, and now the Exception is handled inside `SeedTestData`.

src/com/android/contacts/ui/debug/DebugOptionsMenu.kt
```diff
@@ -0,0 +1,40 @@
...
+    private val seedTestData: SeedTestData,
+    private val clearSeededTestData: ClearSeededTestData,
+) {
+    fun show(context: Context) {
+        val optionsLabels = getOptions().map { context.getString(it.titleId) }.toTypedArray()
+        AlertDialog.Builder(context)
+            .setItems(optionsLabels) { _, which ->
+                CoroutineScope(Dispatchers.Main).launch {
+                    when (DebugOption.entries[which]) {
+                        DebugOption.ExportDatabase -> exportDatabase()
```
RankoR, line 27: If export is unavailable, wrong action will be executed. `optionsLabels` uses filtered version of `DebugOption.entries`, while `when` uses the original indices.

RankoR (changes requested):

src/com/android/contacts/ui/debug/DebugOptionsMenu.kt
sdsantos: Well spotted!

PR #63, opened 2026-08-12
Remove code for SDKs lower than 36

### LINKED ISSUE #64: Remove code for Android SDKs lower than 36

sdsantos:
Closes #64


src/com/android/contacts/model/ContactLoader.java
```diff
@@ -177,23 +177,14 @@ private static class ContactQuery {
                 Data.STATUS_ICON,
                 Data.STATUS_LABEL,
                 Data.STATUS_TIMESTAMP,
+                Data.CARRIER_PRESENCE,
```
m4pl: Looks like this shifts the sequence: everything after it moves by one, but the index constants below still point at the old positions, so `photoUri`, `customRingtone`, `sendToVoicemail` and `isUserProfile` all read the wrong column. Worth appending it at the end of the array instead.

src/com/android/contacts/quickcontact/QuickContactActivity.java
```diff
@@ -2661,10 +2650,7 @@ && isContactEditable()
             ringToneMenuItem.setVisible(!mContactData.isUserProfile() && mArePhoneOptionsChangable);
 
             final MenuItem sendToVoiceMailMenuItem = menu.findItem(R.id.menu_send_to_voicemail);
-            sendToVoiceMailMenuItem.setVisible(
-                    Build.VERSION.SDK_INT < Build.VERSION_CODES.M
-                            && !mContactData.isUserProfile()
-                            && mArePhoneOptionsChangable);
+            sendToVoiceMailMenuItem.setVisible(mArePhoneOptionsChangable);
```
m4pl: The old condition started with `SDK_INT < M`, always false at `minSdk 36`, so this item was always hidden. Now it's visible, and the `!mContactData.isUserProfile()` guard is gone (the ringtone item above still has it). Please check this.

src/com/android/contacts/DynamicShortcuts.java
m4pl: I think `isAtLeastO()` should also be removed, plus `@TargetApi` and `@RequiresApi` annotations.

src/com/android/contacts/compat/telecom/TelecomManagerCompat.java
```diff
@@ -39,10 +38,7 @@ public static void placeCall(@Nullable Activity activity,
         if (activity == null || telecomManager == null || intent == null) {
             return;
         }
-        if (CompatUtils.isMarshmallowCompatible()) {
-            telecomManager.placeCall(intent.getData(), intent.getExtras());
-            return;
-        }
-        activity.startActivityForResult(intent, 0);
+        telecomManager.placeCall(intent.getData(), intent.getExtras());
+        return;
```
m4pl: Think this return; should go.

src/com/android/contacts/compat/PhoneAccountCompat.java
m4pl: `TAG` is unused now that `Log` is gone.

src/com/android/contacts/compat/TelephonyManagerCompat.java
```diff
@@ -17,10 +17,11 @@
...
 
 import android.net.Uri;
-import androidx.annotation.Nullable;
 import android.telecom.PhoneAccountHandle;
 import android.telephony.TelephonyManager;
 
+import androidx.annotation.Nullable;
+
 public class TelephonyManagerCompat {
     public static final String TELEPHONY_MANAGER_CLASS = "android.telephony.TelephonyManager";
```
m4pl: Only existed for the removed `isMethodAvailable`.

src/com/android/contacts/ContactSaveService.java
```diff
@@ -1042,6 +1032,9 @@ public static Intent createSetStarredIntent(Context context, Uri contactUri, boo
     }
 
     private void setStarred(Intent intent) {
+        ContentResolver contentResolver = getContentResolver();
+        if (contentResolver == null) return;
```
m4pl: I think `getContentResolver()` won't return null here.

src/com/android/contacts/model/ContactLoader.java
sdsantos: Ah, missed the indexes below.

src/com/android/contacts/quickcontact/QuickContactActivity.java
sdsantos: You are totally right. But that means this feature disappears from the app, since I found no other flow to manager route/unroute to voicemail. Google's Contacts app maintains this feature. Could have this been a bug?

Related: https://www.reddit.com/r/AndroidQuestions/comments/11nu4iz/is_it_just_me_or_does_the_option_to_route_to/

src/com/android/contacts/ContactSaveService.java
sdsantos: I moved it here because it was checking for null inside `PinnedPositionsCompat.undemote`. But I guess the rest of the code was already trusting it.

m4pl (approved):
LGTM

src/com/android/contacts/ShortcutIntentBuilder.java
```diff
@@ -264,49 +256,27 @@ private Drawable getPhotoDrawable(byte[] bitmapData, String displayName, String
...
-        intent.putExtra(Intent.EXTRA_SHORTCUT_INTENT, shortcutIntent);
-        intent.putExtra(Intent.EXTRA_SHORTCUT_NAME, displayName);
-
+        final long contactId = ContentUris.parseId(contactUri);
+        final ShortcutManager sm = (ShortcutManager)
+                mContext.getSystemService(Context.SHORTCUT_SERVICE);
+        final DynamicShortcuts dynamicShortcuts = new DynamicShortcuts(mContext);
+        final ShortcutInfo shortcutInfo = dynamicShortcuts.getQuickContactShortcutInfo(
+                contactId, lookupKey, displayName);
+        if (shortcutInfo == null) return;
```
RankoR: We shouldn't silently return here. Something like this is better: 

```
if (shortcutInfo == null) {
      mListener.onShortcutIntentCreated(contactUri, null);  
      return;
  }
```

src/com/android/contacts/quickcontact/QuickContactActivity.java
```diff
@@ -2661,10 +2597,8 @@ && isContactEditable()
             ringToneMenuItem.setVisible(!mContactData.isUserProfile() && mArePhoneOptionsChangable);
 
             final MenuItem sendToVoiceMailMenuItem = menu.findItem(R.id.menu_send_to_voicemail);
-            sendToVoiceMailMenuItem.setVisible(
-                    Build.VERSION.SDK_INT < Build.VERSION_CODES.M
-                            && !mContactData.isUserProfile()
-                            && mArePhoneOptionsChangable);
+            sendToVoiceMailMenuItem.setVisible(!mContactData.isUserProfile()
```
RankoR: It was always hidden on pre-M before this change, now it's not

RankoR (changes requested):

src/com/android/contacts/ShortcutIntentBuilder.java
```diff
@@ -326,42 +301,24 @@ private void createPhoneNumberShortcutIntent(Uri uri, String displayName, String
...
-        intent.putExtra(Intent.EXTRA_SHORTCUT_INTENT, shortcutIntent);
-        intent.putExtra(Intent.EXTRA_SHORTCUT_NAME, shortcutName);
+        final IconCompat compatAdaptiveIcon = IconCompat.createWithAdaptiveBitmap(icon);
+        final ShortcutManager sm = (ShortcutManager)
+                mContext.getSystemService(Context.SHORTCUT_SERVICE);
+        final String id = shortcutAction + lookupKey + phoneUri.toString().hashCode();
+        final DynamicShortcuts dynamicShortcuts = new DynamicShortcuts(mContext);
+        final ShortcutInfo shortcutInfo = dynamicShortcuts.getActionShortcutInfo(
+                id, displayName, shortcutIntent, compatAdaptiveIcon.toIcon(mContext));
+        if (shortcutInfo == null) return;
```
RankoR, line 320: Silent return should be fixed here too, missed it on the first review

sdsantos:
I also confirmed the "route to voicemail" is working fine on Android 17 and that it was likely a bug hiding it for Android M+. So now it's going to be available again.

RankoR (approved):

thestinger:
There are conflicts now.

PR #72, opened 2026-08-20
Don't offer local accounts for new labels when default account is cloud

RankoR:
Closes https://github.com/GrapheneOS/os-issue-tracker/issues/8564

### DIFF
--- a/src/com/android/contacts/ContactSaveService.java
+++ b/src/com/android/contacts/ContactSaveService.java
@@ -793,6 +793,7 @@ public class ContactSaveService extends IntentService {
         // added if the group doesn't exist
         if (groupUri == null) {
             Log.e(TAG, "Couldn't create group with label " + label);
+            showToast(R.string.groupCreateFailedToast);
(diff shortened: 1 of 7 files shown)


src/com/android/contacts/ContactSaveService.java
inthewaves: `mGroupsDao.undoDeletion(intent.getBundleExtra(EXTRA_UNDO_DATA));` can now return `null` if undo fails. `mGroupsDao.undoDeletion` will leads to a `insertGroup` call. We should probably do something like `showToast(R.string.groupSavedErrorToast);` if `mGroupsDao.undoDeletion` returns null so user can know if undo fails

inthewaves (approved):
The existing AOSP UX is not ideal (e.g. unlike Google Contacts app, it doesn't clearly show what account a label belongs to), but that's probably beyond the scope of this. The changes LGTM

sdsantos (approved):

PR #78, opened 2026-09-01
Rewrite LicensesActivity to Kotlin + Compose

sdsantos:


src/com/android/contacts/ui/settings/screen/SettingsNavHost.kt
```diff
@@ -57,11 +58,17 @@ internal fun SettingsNavHost(
...
-                        onLicensesClick = { onAction(Action.LicensesClicked) },
+                        onLicensesClick = { route = SettingsNavRoute.Licenses },
                         onNavigateBack = { route = SettingsNavRoute.Main },
                         snackbarHostState = snackbarHostState,
                     )
                 }
+
+                SettingsNavRoute.Licenses -> {
+                    LicensesScreen(
+                        onNavigateBack = { route = SettingsNavRoute.About },
```
RankoR, line 72: Here we're going one level up, but `BackHandler` pops two levels. BackHandler should navigate user to the same level, could be fixed like this:

```
BackHandler(enabled = route != SettingsNavRoute.Main) {
    route = when (route) {
        SettingsNavRoute.Licenses -> SettingsNavRoute.About
        SettingsNavRoute.About, SettingsNavRoute.Main -> SettingsNavRoute.Main
    }
}
```

src/com/android/contacts/ui/settings/about/LicensesScreen.kt
```diff
@@ -0,0 +1,67 @@
...
+        AndroidView(
+            modifier = Modifier
+                .fillMaxSize()
+                .padding(
+                    start = contentPadding.calculateStartPadding(layoutDirection),
+                    end = contentPadding.calculateEndPadding(layoutDirection),
+                    top = contentPadding.calculateTopPadding(),
+                    bottom = contentPadding.calculateBottomPadding(),
+                ),
+            factory = { context ->
```
RankoR, line 50: `onRelease = WebView::destroy` is missing, so WebView is never destroyed

RankoR (changes requested):

src/com/android/contacts/ui/settings/screen/SettingsNavHost.kt
sdsantos: Of course, I caught the button click, but forgot about testing the back press.

m4pl (changes requested):
Could you take a look at https://github.com/GrapheneOS/Messaging/pull/122? There the screen is generated automatically once dependencies are added, so the notices can't drift out of sync with the actual dependency graph.

Unfortunately there's nothing about license generation in https://github.com/GrapheneOS/app-docs.

RankoR (approved):

PR #83, opened 2026-09-02
Contact Details: Java/XML to Kotlin/Compose migration

### LINKED ISSUE #55: Rewrite contact details to Compose

m4pl:
Added:
- Recent calls section
- SIM selection for calls
- Editing a number from the number menu
- Send to voicemail toggle
- Connected apps section
- Navigation to group members from a contact
- Source account of the contact in the UI

### Screenshots

| Before | After |
|--------|-------|
|Screenshot_20260903_122740|Screenshot_20260903_122949|

### Video
[Video attachment](https://github.com/user-attachments/assets/99205271-0486-467c-95c1-33852348326f)

Closes #55


m4pl:
Also, here is the generated VCF file:
[Sample.vcf.zip](https://github.com/user-attachments/files/31789162/Sample.vcf.zip)

src/com/android/contacts/di/contactdetails/ContactDetailsBindsModule.kt
```diff
@@ -0,0 +1,148 @@
...
+    ): ContactEntryIntentFactory
+
+    @Binds
+    @Reusable
+    abstract fun bindContactLoaderSource(
+        impl: ContactLoaderSourceImpl,
+    ): ContactLoaderSource
+
+    @Binds
+    @Singleton
```
RankoR: Repository contains per-screen mutable state - so it can't be a singleton. Or, better, repository should not contain per-screen state.

src/com/android/contacts/ui/contactdetails/screen/ContactDetailsViewModel.kt
```diff
@@ -0,0 +1,387 @@
...
+    private val _navigationEvents = Channel<NavEvent>(Channel.BUFFERED)
+    override val navigationEvents: Flow<NavEvent> = _navigationEvents.receiveAsFlow()
+
+    private val callingSimPickerVisible = MutableStateFlow(false)
+
+    private val argumentsBundle: StateFlow<Bundle?> = savedStateHandle
+        .getStateFlow(KEY_ARGUMENTS, null)
+
+    private val arguments: Flow<ContactDetailsArguments> = argumentsBundle
+        .mapNotNull { bundle -> bundle?.toArguments() }
```
RankoR, line 82: Needs `.distinctUntilChanged()`, otherwise it re-launches whole pipeline on each config change (rotation, etc)

src/com/android/contacts/ui/contactdetails/ContactDetailsActivity.kt
```diff
@@ -0,0 +1,139 @@
...
+            finish()
+        }
+    }
+
+    private fun applyDirectoryCopyResult(result: ActivityResult) {
+        setResult(result.resultCode)
+
+        val intent = result.toRebindIntent() ?: return
+
+        bindIntent(intent)
```
RankoR, line 122: Need `setIntent(intent)`  before `bindIntent`, like in `onNewIntent`.

RankoR (changes requested):

RankoR (approved):

src/com/android/contacts/data/contactdetails/repository/ContactActionsRepository.kt
```diff
@@ -0,0 +1,189 @@
...
+                ContentUris.withAppendedId(Data.CONTENT_URI, dataId),
+                values,
+                null,
+                null,
+            )
+        }
+    }
+
+    override fun observeLinkOperations(): Flow<ContactLinkOperation> {
+        return callbackFlow {
```
sdsantos, line 138: There's the `BuildBroadcastReceiverFlow`.

src/com/android/contacts/ui/common/components/ContactAvatarColors.kt
```diff
@@ -0,0 +1,145 @@
...
+
+    forEach { character ->
+        hash = hash xor character.code
+        hash *= FNV_PRIME
+    }
+
+    return hash
+}
+
+private fun hslColor(
```
sdsantos: There is a `Color.hsl(...)` method already. Is this implementation different?

sdsantos:
I'm still half-way through the code review. In the meantime, here are the UI issues I found:

a) Video quick action is always disabled, even on my regular phone (where it is enabled on google contacts).

b) The recent activity cells right padding looks to be a bit small.

Screenshot 2026-09-09 at 17 22 25

c) The connected apps title appearance looks very similar with the description, maybe some bold is missing from the apps title?

Screenshot 2026-09-09 at 17 22 45

d) Directions intent just opening location, and not starting directions.

e) Sometimes setting a different ringtone fails to update the UI, and I need to repeat the action.

f) The default empty avatar color does not match the one in the contacts list (I imagine it will in the future?).

g) Account info looks a bit off using ContactAvatar.

Screenshot 2026-09-09 at 18 56 29

h) In landspace, the contact name is clipped and the nicknames don't appear:

Screenshot_20260909_174526
Screenshot_20260909_174512

src/com/android/contacts/ui/contactdetails/screen/ContactDetailsCards.kt
```diff
@@ -0,0 +1,161 @@
...
+) {
+    val headerHeight = rememberOverlayHeight()
+    val quickActionsHeight = rememberOverlayHeight()
+    val isListLaidOut by remember(listState) {
+        derivedStateOf { listState.layoutInfo.totalItemsCount > 0 }
+    }
+    val overlayAlpha = when {
+        isListLaidOut -> VISIBLE_ALPHA
+        else -> HIDDEN_ALPHA
+    }
```
sdsantos: I'm curious why did you go with this approach versus everything inside a `LazyColumn` with the QuickActions as a `stickyHeader`. Is there any issue with that approach that I missed?

src/com/android/contacts/ui/contactdetails/screen/ContactDetailsViewModel.kt
```diff
@@ -0,0 +1,397 @@
...
+            arguments = arguments,
+        )
+
+        linkDelegate.unlinked
+            .onEach { sendNavigationEvent(NavEvent.Close) }
+            .launchIn(viewModelScope)
+
+        contentDelegate.loadedDetails
+            .filterNotNull()
+            .onEach { details -> reportShortcutUsed(details) }
```
sdsantos, line 122: We currently have no way of telling if the user came from a shortcut right? That's why we're always reporting it?

sdsantos (changes requested):
Did not find anything else in terms of UI issues.

src/com/android/contacts/ui/common/components/ContactAvatarColors.kt
m4pl: Same conversion indeed, switched to `Color.hsl`.

src/com/android/contacts/data/contactdetails/repository/ContactActionsRepository.kt
m4pl: It only emits `Unit`, and here we need the intent action. It also replays on subscribe, which we don't want.

src/com/android/contacts/ui/contactdetails/screen/ContactDetailsViewModel.kt
m4pl: Right, we can't tell.

src/com/android/contacts/ui/contactdetails/screen/ContactDetailsCards.kt
m4pl: You're right, no reason. Reworked. Also fixed the landscape clipping.

m4pl:
@sdsantos 

**(a) Video kept as is:** this matches AOSP, where the action needs a phone account with `CAPABILITY_VIDEO_CALLING`. Google Contacts enables it through Duo/Meet, which has no counterpart here.
**(f) Avatar kept as is:** list is still on `LetterTileDrawable` with its own palette. It will match once it moves to `ContactAvatar`.

src/com/android/contacts/ui/common/components/AccountIcon.kt
```diff
@@ -8,30 +12,70 @@ import androidx.compose.ui.layout.ContentScale
...
+        iconData.titleRes,
+        iconData.iconRes,
+        iconData.syncAdapterPackageName,
     )
+
+    return when {
+        icon == null || !iconData.applyGrayTint -> icon
+
+        else -> icon.mutate().apply {
+            colorFilter = BlendModeColorFilter(
```
sdsantos, line 73: We can also use the compose argument `colorFilter` directly on `Image`:
```
colorFilter = BlendModeColorFilter(colorResource(R.color.actionbar_icon_color_grey), BlendMode.SrcAtop),
```

sdsantos (approved):
Just some smaller UI issues:
1. Directions button now starts navigation straight away, instead of just showing navigation options like Google Contacts does.
2. Setting a Ringtone as None versus Default is not distinguishable in the details cell.

m4pl:
> Directions button now starts navigation straight away, instead of just showing navigation options like Google Contacts does

@sdsantos I made it intentionally. The only way to get the options screen is via `https://www.google.com/maps/dir`, which I'd rather avoid. @RankoR what do you think?

PR #84, opened 2026-09-08
Rewrite ShowOrCreateActivity to Kotlin + Compose

### LINKED ISSUE #82: Rewrite ShowOrCreate to Kotlin + Compose

sdsantos:
Closes #82


src/com/android/contacts/ui/interactions/showorcreate/ShowOrCreateActivity.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.android.contacts.ui.interactions.showorcreate.screen.ShowOrCreateDialog
+import com.android.contacts.ui.interactions.showorcreate.screen.ShowOrCreateEffectHandlerImpl
+import com.android.contacts.ui.interactions.showorcreate.screen.ShowOrCreateViewModel
+import dagger.hilt.android.AndroidEntryPoint
+
+@AndroidEntryPoint
+class ShowOrCreateActivity : FragmentActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
```
RankoR, line 18: Old activity checked for permission, current one does not. We should handle missing permission, at least with `if (RequestPermissionsActivity.startPermissionActivityIfNeeded(this)) return`

src/com/android/contacts/ui/interactions/showorcreate/screen/ShowOrCreateEffectHandler.kt
```diff
@@ -0,0 +1,52 @@
...
+internal class ShowOrCreateEffectHandlerImpl(
+    private val activity: Activity,
+) : ShowOrCreateEffectHandler {
+    override fun handle(effect: Effect) {
+        when (effect) {
+            Effect.Close -> {
+                activity.finish()
+            }
+
+            is Effect.CreateContact -> {
```
RankoR: Old activity had two different intents:

- Force create: `ACTION_INSERT` + `RawContacts.CONTENT_TYPE`
- `ACTION_INSERT_OR_EDIT` + `RawContacts.CONTENT_ITEM_TYPE`

Need to restore it, by adding a separate `Effect.CreateOrEditContact` -> `ACTION_INSERT_OR_EDIT` + `CONTENT_ITEM_TYPE`

src/com/android/contacts/data/contacts/repository/ContactsRepository.kt
```diff
@@ -0,0 +1,133 @@
...
+            failedLookup(e)
+        }
+    }
+
+    private fun toLookupResults(cursor: Cursor): List<ContactLookupResult> {
+        val results = mutableListOf<ContactLookupResult>()
+        while (cursor.moveToNext()) {
+            val id = cursor.getLong(LOOKUP_ID_INDEX)
+            val key = cursor.getString(LOOKUP_KEY_INDEX)
+            results.add(
```
RankoR: `key` == `null` or `uri` == `null` will crash the app.

```
   while (cursor.moveToNext()) {
        val id = cursor.getLong(LOOKUP_ID_INDEX)
        val key = cursor.getString(LOOKUP_KEY_INDEX) ?: continue
        val uri = ContactsContract.Contacts.getLookupUri(id, key) ?: continue
        results.add(...)
    }
```

app/src/test/kotlin/com/android/contacts/ui/interactions/showorcreate/screen/ShowOrCreateViewModelTest.kt
```diff
@@ -0,0 +1,218 @@
...
+class ShowOrCreateViewModelTest {
+
+    @get:Rule
+    val mainDispatcherRule = MainDispatcherRule()
+
+    private val contactsRepository = mockk<ContactsRepository>(relaxed = true)
+
+    @Before
+    fun setUp() {
+        mockkStatic(Uri::class)
```
RankoR, line 42: This is never unmocked

src/com/android/contacts/ui/interactions/showorcreate/screen/ShowOrCreateEffectHandler.kt
```diff
@@ -0,0 +1,52 @@
...
+import android.provider.ContactsContract
+import com.android.contacts.activities.PeopleActivity
+import com.android.contacts.ui.interactions.showorcreate.screen.model.ShowOrCreateEffect as Effect
+import com.android.contacts.util.ImplicitIntentsUtil
+
+internal interface ShowOrCreateEffectHandler {
+    fun handle(effect: Effect)
+}
+
+internal class ShowOrCreateEffectHandlerImpl(
```
RankoR, line 15: Should be covered by `ShowOrCreateEffectHandlerImplTest`.

app/src/test/kotlin/com/android/contacts/ui/interactions/showorcreate/screen/ShowOrCreateViewModelTest.kt
```diff
@@ -0,0 +1,218 @@
...
+            }
+            val subject = createViewModel(
+                savedState = mapOf(
+                    ShowOrCreateViewModel.EXTRA_DATA to "mailto:$email".toUri(),
+                    ShowOrCreateViewModel.EXTRA_EXTRAS to bundle,
+                ),
+            )
+            subject.effects.test {
+                advanceUntilIdle()
+                assertEquals(
```
RankoR, line 136: This checks only the class, but not it's contents, so wrong bundle contents will be missed.
Same below.

src/com/android/contacts/ui/interactions/showorcreate/screen/ShowOrCreateViewModel.kt
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+        query?.let(::lookupContacts)
+    }
+
+    override fun onAction(action: Action) {
+        when (action) {
+            Action.CreateConfirm -> {
+                val query = query ?: return
+                emitEffect(Effect.CreateContact(buildCreateExtras(query)))
```
RankoR: Double tap can emit two `CreateContact` effects

src/com/android/contacts/data/contacts/repository/ContactsRepository.kt
```diff
@@ -0,0 +1,133 @@
...
+                    key = key,
+                    uri = ContactsContract.Contacts.getLookupUri(id, key),
+                ),
+            )
+        }
+        return results
+    }
+
+    private fun failedLookup(cause: Exception): List<ContactLookupResult> {
+        Log.w(TAG, "Could not load the user profile", cause)
```
RankoR: Message seems to be wrong

src/com/android/contacts/ui/interactions/showorcreate/screen/ShowOrCreateDialog.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    ShowOrCreateDialogContent(
+        uiState = uiState,
+        onAction = screenModel::onAction,
+        modifier = modifier,
+    )
+}
+
+@Composable
+private fun ShowOrCreateDialogContent(
```
RankoR, line 41: No preview for this composable

RankoR (changes requested):

AndroidManifest.xml
```diff
@@ -288,9 +288,9 @@
             android:theme="@style/PeopleActivityTheme"/>
 
         <activity
-            android:name=".activities.ShowOrCreateActivity"
+            android:name=".ui.interactions.showorcreate.ShowOrCreateActivity"
             android:exported="true"
```
m4pl, line 292: The component name changed. Worth an `<activity-alias>` under the old name.

res/values/styles.xml
```diff
@@ -20,6 +20,13 @@
         <item name="alertDialogTheme">@style/ContactsAlertDialogThemeAppCompat</item>
     </style>
 
+    <style name="Theme.Compose.Dialog" parent="Theme.Compose">
+        <item name="android:windowBackground">@android:color/transparent</item>
+        <item name="android:windowContentOverlay">@null</item>
```
m4pl, line 25: `Theme.Material.Light` already sets `windowContentOverlay`.

src/com/android/contacts/ui/interactions/showorcreate/screen/ShowOrCreateViewModel.kt
```diff
@@ -0,0 +1,136 @@
...
+) : ViewModel(),
+    ShowOrCreateScreenModel {
+
+    private val _effects = Channel<Effect>(capacity = Channel.BUFFERED)
+    override val effects: Flow<Effect> = _effects.receiveAsFlow()
+
+    private val _uiState = MutableStateFlow<State>(State.Searching)
+    override val uiState = _uiState.asStateFlow()
+
+    private val data: Uri?
```
m4pl, line 48: These come from the intent and cannot change, so a `val` read once would say that. Same for `query` at line `62`, which is only assigned in `init`.

src/com/android/contacts/ui/interactions/showorcreate/screen/ShowOrCreateViewModel.kt
```diff
@@ -0,0 +1,136 @@
...
+                    emitEffect(Effect.CreateContact(buildCreateExtras(query)))
+                }
+                else -> {
+                    _uiState.value = State.ConfirmingCreate(createDescription)
+                }
+            }
+        }
+    }
+
+    private fun buildCreateExtras(query: ContactLookupQuery): Bundle {
```
m4pl, line 118: If the effects carried the `ContactLookupQuery` instead, the handler could build the bundle from `activity.intent.extras` and `EXTRA_EXTRAS` would not be needed.

src/com/android/contacts/ui/interactions/showorcreate/screen/model/ShowOrCreateEffect.kt
```diff
@@ -0,0 +1,24 @@
...
+
+internal sealed interface ShowOrCreateEffect {
+    data object Close : ShowOrCreateEffect
+
+    data class ShowContact(
+        val uri: Uri,
+    ) : ShowOrCreateEffect
+
+    data class ShowContactList(
+        val extras: Bundle,
```
m4pl, line 14: `Bundle` does not override `equals`, so these data classes compare by reference. `ShowOrCreateViewModelTest.kt:232` has to stub `deepCopy()` to return the same mock so `assertEquals` passes.

src/com/android/contacts/data/contacts/repository/ContactsRepository.kt
```diff
@@ -0,0 +1,139 @@
...
+        return when (query) {
+            is ContactLookupQuery.Email ->
+                Uri.withAppendedPath(
+                    ContactsContract.CommonDataKinds.Email.CONTENT_FILTER_URI,
+                    Uri.encode(query.value),
+                )
+            is ContactLookupQuery.Phone -> {
+                Uri.withAppendedPath(
+                    ContactsContract.PhoneLookup.CONTENT_FILTER_URI,
+                    Uri.encode(query.value),
```
m4pl, line 63: The old code appended the phone number without `Uri.encode`, email with. Encoding both looks right for numbers containing `#`, just confirming it is intentional.

src/com/android/contacts/data/contacts/repository/ContactsRepository.kt
```diff
@@ -0,0 +1,139 @@
...
+                getQueryUri(query),
+                getQueryProjection(query),
+                null,
+                null,
+                null,
+            )
+                ?.use(::toLookupResults)
+                ?: emptyList()
+        } catch (e: SecurityException) {
+            failedLookup(e)
```
m4pl, line 88: A failed lookup returns `emptyList()`, so the ViewModel cannot tell it apart from no matches and shows the create prompt. The old path went through `AsyncQueryHandler`, which turns any exception into a null cursor and made the activity `finish()`. Should failures be a distinct result?

src/com/android/contacts/ui/interactions/showorcreate/screen/ShowOrCreateViewModel.kt
```diff
@@ -0,0 +1,136 @@
...
+        query = when (scheme) {
+            ContactsUtils.SCHEME_MAILTO -> {
+                ContactLookupQuery.Email(schemeSpecificPart.orEmpty())
+            }
+            PhoneAccount.SCHEME_TEL -> {
+                ContactLookupQuery.Phone(schemeSpecificPart.orEmpty())
+            }
+            else -> {
+                Log.w(TAG, "Invalid intent scheme: $scheme")
+                emitEffect(Effect.Close)
```
m4pl, line 74: This branch closes the screen and returns `null` only to fill the expression, so the `when` both maps the scheme and aborts. An early return would split them, and `query` could stop being nullable.

src/com/android/contacts/ui/interactions/showorcreate/ShowOrCreateActivity.kt
```diff
@@ -0,0 +1,40 @@
...
+        enableEdgeToEdge()
+
+        if (RequestPermissionsActivity.startPermissionActivityIfNeeded(this)) {
+            return
+        }
+
+        intent.putExtra(ShowOrCreateViewModel.EXTRA_DATA, intent.data)
```
m4pl, line 25: Line `25` adds `data` to the extras, so the bundle line `26` stores already contains it, and that stray key ends up in the insert intent. The intent also survives recreation, so the bundle nests one level deeper on every rotation.

m4pl (changes requested):

src/com/android/contacts/data/contacts/repository/ContactsRepository.kt
sdsantos: It was not intentional, but let's leave it then 😅

src/com/android/contacts/ui/interactions/showorcreate/screen/ShowOrCreateViewModel.kt
sdsantos: Not sure how do you mean to return out of an `init` or value initialization, to avoid it being nullable.

m4pl: My mistake, query stays nullable. I meant this:
```
private val query = when (scheme) {
    ContactsUtils.SCHEME_MAILTO -> ContactLookupQuery.Email(schemeSpecificPart.orEmpty())
    PhoneAccount.SCHEME_TEL -> ContactLookupQuery.Phone(schemeSpecificPart.orEmpty())
    else -> null
}

init {
    if (query == null) {
        Log.w(TAG, "Invalid intent scheme: $scheme")
        emitEffect(Effect.Close)
    } else {
        lookupContacts(query)
    }
}
```

